PR #78, opened 2026-03-02
feat: fuzzy search for branch selection prompts

KacperKazan:
Replace `dialoguer::Select` with `FuzzySelect` in all branch selection prompts so users can type to filter branches instead of only using arrow keys.

**Affected commands:**
- `stax branch reparent` — parent branch selection
- `stax branch track` — parent branch selection
- `stax up` / `stax top` — child branch selection
- `stax init` — trunk branch selection


cesarferreira (approved):
really cool

PR #126, opened 2026-03-18
fix: not being able to delete branch

fabis94:
## Summary

Fix branch deletion failing for branches with slashes in their names (e.g. `feature/my-branch`).

### Branch deletion via git CLI

Replaced `libgit2`'s `branch.delete()` with a direct `git branch -d/-D` CLI call. The libgit2 method fails to clean up branch config entries when the branch name contains slashes, which caused deletion to error out or leave the repo in a broken state.

### Restructured force-delete logic

Reorganized the control flow so the merged-into-trunk safety check only runs when `force` is false, then deletion always goes through the CLI path regardless of the force flag.

## How it works

- **Behavior**: `stax branch delete` (and any other delete paths) now correctly removes branches with slash-separated names without errors.
- **Implementation**: After the optional safety check, deletion is delegated to `git branch -d` or `git branch -D` via the existing `run_git` helper. stderr is captured and surfaced as an error if the command fails.
- **Tradeoffs**: Delegates to the git CLI rather than the libgit2 API, which is consistent with how other operations in this codebase handle cases where libgit2 falls short. Requires git to be available on PATH (already a hard requirement elsewhere).

## Configuration changes

- [ ] No config changes
- [ ] Config changes included (describe below)

## Testing

- [ ] `make test` (or `just test`) for full-suite validation
- [ ] `cargo check`
- [ ] `cargo clippy -- -D warnings`
- [ ] `cargo fmt -- --check`
- [ ] Manual verification (if applicable)

Commands/output (if relevant):

```bash
# Reproduce the fix:
# 1. Create a branch with a slash in the name
git checkout -b feature/test-branch
# 2. Switch away and attempt deletion
git checkout main
stax branch delete feature/test-branch
# Previously errored; now deletes cleanly
```


cesarferreira:
@fabis94 sorry it was claude that wrote that.
if you fix the conflicts i can merge ;)

fabis94:
fixed 👍

Issue #200, opened 2026-04-04
sync --delete-upstream-gone can orphan child branches by deleting parent metadata without reparenting


cesarferreira:
## Problem
`stax sync --delete-upstream-gone` deletes a stale local branch, but it does **not** reparent any tracked children first.

That leaves the stack metadata in an inconsistent state: descendants can still point at the deleted parent, and `stax status --json` reports the child with `"parent": "<deleted-branch>"` and `needs_restack: false`.

## Repro
In a fresh repo with a remote:

```bash
stax init --trunk main
stax create parent
# commit + push parent
stax create child
# commit + push child

git checkout main
git push origin --delete parent

stax sync --force --delete-upstream-gone
stax status --json
```

Observed after the sync:
- `parent` is deleted locally
- `child` remains tracked
- `child` still points to `parent` in stax metadata even though `parent` no longer exists
- `child` is not marked for restack

I reproduced this locally against `v0.41.0`.

## Likely affected area
- `src/commands/sync.rs`
  - the `delete_upstream_gone` path deletes the branch, but unlike merged-branch cleanup it does not reparent children or update branch metadata first

## Proposed improvement
Make `sync --delete-upstream-gone` mirror the merged-branch cleanup flow before deleting a tracked branch:
- find tracked children whose parent is the upstream-gone branch
- reparent them to a sensible fallback (recorded parent if it still exists, otherwise trunk)
- update metadata so descendants become restackable / consistent
- add an integration test that covers `parent -> child`, deleting the parent's upstream, and verifying the child no longer references a deleted branch

## Expected benefit
Prevents orphaned stacks after routine cleanup, keeps `stax status` truthful, and makes `sync --delete-upstream-gone` safe for tracked stacked branches instead of only for leaf branches / untracked branches.

Issue #198, opened 2026-04-03
sync should not treat closed-but-unmerged PRs as merged branches


cesarferreira:
## Summary
`stax sync` currently classifies any tracked branch with `pr_state == "closed"` as a merged branch candidate, even if the PR was closed without merge and the branch still exists locally/remotely.

## Evidence
In `src/commands/sync.rs`, `find_merged_branches()` adds a branch to the merged-cleanup list when:

```rust
if matches!(
    info.pr_state.as_deref(),
    Some(state)
        if state.eq_ignore_ascii_case("merged") || state.eq_ignore_ascii_case("closed")
) {
    merged.push(branch.clone());
}
```

That means a branch can be surfaced for deletion purely because its PR is `closed`, regardless of whether:
- the branch tip is actually merged into trunk
- the remote branch still exists
- the branch was intentionally kept after closing the PR

Later in the same sync flow, cleanup can force-delete the local branch and also tries to delete the remote branch.

## Affected area
- `src/commands/sync.rs`
- specifically `find_merged_branches()` and the merged-branch cleanup path

## Proposed improvement
Only treat PR state `merged` as a direct cleanup signal.
For `closed` PRs, require an additional proof of merge/deletion before offering cleanup, for example:
- remote branch is gone, or
- branch is reachable/patch-equivalent to trunk, or
- explicit `--delete-closed-pr-branches` style opt-in if that behavior is still desired

## Expected benefit
This avoids accidental cleanup of branches whose PRs were closed without merge, especially on `stax sync --force`, and makes the "delete merged branches" step match the label more accurately.

PR #208, opened 2026-04-04
fix: don't treat closed-but-unmerged PRs as merged during sync

cesarferreira:
## Summary
- stop `sync` from treating PR metadata with state `closed` as if the branch were merged
- keep only explicit `merged` metadata as a direct cleanup signal
- add an integration test covering a tracked branch with closed PR metadata that still has unmerged work

## Notes
- I found issue #200 currently describes a different `--delete-upstream-gone` bug, so this PR follows the Kanban task title / code path instead
- I could not run the targeted Rust test on this host because `openssl-sys` fails to build here (`openssl.pc` / OpenSSL dev files are missing on the Pi environment)

Refs #200

### DIFF
--- a/src/commands/sync.rs
+++ b/src/commands/sync.rs
@@ -1277,7 +1277,9 @@ fn find_merged_branches(
         }
     }
 
-    // Method 2: Check PR state from metadata - if PR is merged, branch should be deleted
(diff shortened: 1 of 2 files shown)


cesarferreira:
Superseded by #207; same fix, but #207 is aligned to issue #198 and has the cleaner regression test implementation.

PR #296, opened 2026-04-16
Improve st setup onboarding flow

cesarferreira:
## Motivation

`st setup` should be the one-shot onboarding command for a new machine, rather than only shell integration. This change teaches setup to handle shell integration, optional agent skills, and GitHub auth onboarding in one flow, and updates the docs to present it that way.

- Extend `st setup` into a broader onboarding flow with explicit flags:
- `--yes` to accept default setup actions
- `--install-skills` / `--skip-skills`
- `--auth-from-gh` / `--skip-auth`
- Keep `--print` and `--refresh` for the existing shell-setup paths

- Reuse existing commands instead of duplicating logic:
- setup still owns shell integration and rerere enablement
- skills install delegates to `stax skills update`
- auth import delegates to `stax auth --from-gh` behavior

- New auth behavior in setup:
- if auth is already configured, setup leaves it alone
- if `gh auth token` is available, setup can prompt to import it
- if `--yes` is used and `gh` is unavailable, setup prints next steps instead of dropping into an interactive token prompt

- Update README/docs/skills guidance so `st setup` is described as the one-shot onboarding entry point, with `st setup --yes` shown as the “do it all” path

- Add CLI coverage for:
- skills/auth flag conflicts
- interactive accept/decline paths
- `--yes` installing skills and importing auth from `gh`
- `--yes` printing auth next steps when `gh` is unavailable
- portable interactive shell tests that now run against bash in CI instead of assuming `/bin/zsh` exists

## Testing

- `cargo test --test cli_tests shell_setup -- --nocapture`


cesarferreira:
## Stack Links

* `main`
  * **PR #296** 👈

Issue #312, opened 2026-04-20
fix(merge): only retarget remaining PRs after the rewritten branch has been pushed successfully


cesarferreira:
After a successful rebase of remaining branches, both merge flows retarget PR bases before confirming the force-push succeeded, and then ignore the push result.

Evidence:
- `src/commands/merge.rs:438-451`
- `src/commands/merge_when_ready.rs:448-461`
- both paths call `client.update_pr_base(...)` first
- the following `git push --force-with-lease ...` result is discarded with `let _ = ...output()`

Why this is risky:
- if the push is rejected (stale lease, auth problem, transient network issue, branch protection, etc.), the forge PR base has already changed while the remote branch tip has not
- that can leave the stack in a misleading half-updated state, with PR relationships reflecting a rewrite that never actually reached the remote
- this is the same class of ordering problem that previously caused retarget/merge inconsistencies in the merge path, just in the remaining-branch phase

Suggested fix:
- only update the PR base after the force-push succeeds
- surface push failures instead of silently continuing
- add regression coverage that simulates a push failure after a successful local rebase and verifies the PR base is left unchanged

PR #318, opened 2026-04-21
fix(merge-when-ready): preserve remaining stack chain on rebase (#311)

### LINKED ISSUE #311: fix(merge-when-ready): preserve the remaining stack chain instead of flattening everything onto trunk
`st merge --when-ready` currently rebases every remaining branch directly onto trunk after the lower part of the stack is merged.

Evidence:
- `src/commands/merge_when_ready.rs:430-462` loops over `remaining_branches`
- inside that loop it always calls `rebase_descendant_onto_remote_trunk_with_provenance(..., &scope.trunk, ...)`
- it also updates every remaining PR base to `scope.trunk`

That differs from `st merge`, which now preserves the relative chain of the remaining stack by rebasing the first remaining branch onto trunk and each later branch onto the previous remaining branch (`src/commands/merge.rs:396-454`).

Why this matters:
- for a stack like `main <- a <- b <- c <- d`, merging `a` and `b` with `st merge --when-ready` should leave `c` based on `main` and `d` based on `c`
- the current code appears to flatten both `c` and `d` onto `main`
- that changes the PR topology and can produce much larger diffs for higher branches

Suggested fix:
- mirror the `st merge` remaining-branch logic in `merge_when_ready`
- preserve the branch chain when rebasing and retargeting PR bases
- add an integration test that merges a partial stack and verifies the top branch still points to the previous remaining branch, not trunk

geoHeil:
## Summary
`merge --when-ready` rebased every remaining descendant onto trunk and updated each PR base to trunk, flattening stacks of the form `main <- a <- b <- c <- d` into `main <- c` + `main <- d` whenever `a` and `b` merged. The regular `merge` already walked the remaining list chain-wise; mirror that logic so the topology (and the PR diff size) of higher branches stays stable.

- Walk remaining branches chain-wise in `merge --when-ready` (first → trunk, then each onto previous remaining)
- Extract `rebase_and_finalize_remaining_branch` shared helper; both merge flows now share the full fetch + rebase + push + retarget sequence
- Add integration test asserting \`PR_d.base == branch_c\` (not flattened to trunk)

Fixes #311.

**Note:** stacked on top of #312 (#317). Rebase onto main once #312 merges.

## Test plan
- [x] `make test` — full suite passes (Docker Linux)
- [x] New test: `test_merge_when_ready_preserves_remaining_chain` builds a 4-branch stack, merges the lower two, and verifies remaining PRs retarget to their previous remaining branch rather than flattening onto trunk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+- Merge (when-ready): Preserve the relative stack chain when rebasing remaining branches instead of flattening every descendant onto trunk (#311)
+
 ## [0.56.0] - 2026-04-21
 
 ### Changed
--- a/src/commands/merge.rs
+++ b/src/commands/merge.rs
@@ -396,70 +396,22 @@ pub fn run(
         }
 
         for (idx, remaining) in scope.remaining.iter().enumerate() {
-            let parent_branch = if idx == 0 {
-                scope.trunk.clone()
+            let previous = if idx == 0 {
+                None
             } else {
-                scope.remaining[idx - 1].branch.clone()
+                Some(scope.remaining[idx - 1].branch.as_str())
             };
-            let parent_is_trunk = idx == 0;
-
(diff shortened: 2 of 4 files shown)


cesarferreira:
thanks for the PR! your previous PR made conflicts with this one, can you rebase please?

PR #342, opened 2026-04-27
fix(status): avoid slow ls git scans

cesarferreira:
## Summary

Fixes slow first-run `st ls`/`st status` in large repositories by removing broad repository scans from the default status path and by parallelizing the branch graph work that still has to happen.

## Root cause

The original regression was not a stale install. I verified the installed `st` binary was the local build from this branch, then reproduced the remaining delay in `wayve/frontends/robot-android`.

There were three separate costs in the status path:

1. `st ls` used to shell out for broad repo-wide data even though the repo was already open through libgit2:
   - `git branch -r --format=%(refname)` to enumerate every remote branch
   - `git worktree list --porcelain` to discover linked worktrees
   - `git rev-parse --git-dir` for rebase-state checks
2. The first fix moved those reads to libgit2, but the remote branch check was still broad: it iterated every remote branch just to decide whether the handful of displayed stack branches should show the cloud marker.
3. In the large robot-android repo, the remaining visible delay came from serial ahead/behind graph walks. The stack in the report has five displayed branches, and each branch compares against `main`, producing counts such as `1179 behind 1 ahead`. Running those graph walks one after another made cold `st ls` feel like a hang.

## What changed

- Replaced status/list remote detection with targeted libgit2 ref lookups for only the branches being rendered.
- Replaced status/list worktree detection with a lightweight libgit2 worktree map instead of `git worktree list --porcelain`.
- Replaced rebase-state subprocess discovery with direct checks under the already-open repo git dir.
- Added `GitRepo::commits_ahead_behind_many`, which opens independent repo handles and computes the required ahead/behind pairs concurrently while preserving output order.
- Kept JSON and text output semantics unchanged: missing or unresolvable counts still degrade to `0 ahead / 0 behind`, matching the previous best-effort behavior.
- Added focused regression coverage so `st ls` keeps avoiding the old broad subprocess scans and so the new batch ahead/behind helper matches the existing single-pair implementation.

## Timing notes

Direct checks in `wayve/frontends/robot-android` after installing this branch:

- `make install` replaced `/Users/cesarferreira/.cargo/bin/st` with the local debug build from this branch.
- Immediate first run after overwriting the installed debug binary: `real 2.48s`.
- Repeated installed run: `real 0.36s`.
- Fresh copied debug binary first run, which avoids the install overwrite noise but still exercises a cold binary path: `real 0.98s`; repeated run: `real 0.36s`.

Before this follow-up fix, the same large repo was still taking about `2.6s` directly from the binary after the first PR version, and the user's shell prompt reported `6s` for the first interactive `st ls`. The remaining delay was therefore real code-path work, not an old binary.

## Verification

- `cargo build --quiet --bins`
- `cargo build --release --quiet --bins`
- `cargo test test_commits_ahead_behind_many_matches_single_queries --lib -- --nocapture`
- `cargo test test_get_existing_remote_branches_from_repo_checks_only_requested_branches --lib -- --nocapture`
- `cargo test status_uses_cached_repo_data_without_git_subprocess_scans --test worktree_tests -- --nocapture`
- `cargo test test_status --test comprehensive_coverage_tests -- --nocapture`
- `cargo test test_status_during_rebase_gives_clear_error --test conflict_handling_tests -- --nocapture`
- `git diff --check`
- Manual large-repo timing in `wayve/frontends/robot-android` with `st ls --quiet`

## Docs

No README/docs/skills command docs were updated because this does not change user-visible command syntax, flags, output shape, or defaults. It is an internal performance fix for existing `st ls`/`st status` behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
 One giant PR is slow to review and risky to merge. A stack of small PRs is the answer — but managing stacks by hand with `git rebase --onto` is a footgun. **stax** makes stacks a first-class Git primitive.
 
 - **Stack, don't wait.** Keep shipping on top of in-review PRs. `st create`, `st ss`, done.
-- **Native-fast.** A single Rust binary that starts in ~25ms. `st ls` benches ~16× faster than Graphite/Freephite on this repo.
+- **Native-fast.** A single Rust binary that starts in ~25ms. `st ls` benches ~70× faster than Graphite and ~215× faster than Freephite on this repo.
 - **Agent-native.** Run parallel AI agents on isolated branches (`st lane`), auto-resolve rebase conflicts (`st resolve`), and generate PR bodies from real diffs.
 - **Undo-first.** Every destructive op snapshots state. `st undo` / `st redo` rescue risky rebases instantly.
 - **Batteries-included TUI.** Run bare `st` to browse the stack, inspect diffs, and watch CI hydrate live.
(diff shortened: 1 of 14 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #342** 👈

PR #348, opened 2026-04-29
test(refresh): cover squash-merged parent restack

cesarferreira:
## Summary

- Add regression coverage proving `st refresh --no-submit --force` handles a squash-merged middle branch through the same provenance-aware sync restack path as `st rs --restack`.
- Assert the merged parent branch is cleaned up and the child keeps only its own novel commit over `main`.

## Test Plan

- `cargo test --test integration_tests test_refresh -- --nocapture`
- `cargo fmt --check`
- `git diff --check HEAD`

## Docs

- Not updated; this is test-only regression coverage with no user-visible behavior change.

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -2205,6 +2205,80 @@ fn test_refresh_no_submit_keeps_original_branch_and_restacks_stack() {
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #348** 👈

PR #360, opened 2026-05-06
feat: support repo-local config

gabriel-ecegi:
## Summary
- resolve .config/stax/config.toml from the current git repository root when present
- keep STAX_CONFIG_DIR as the highest-priority override and global config as fallback
- document config precedence and repo-local worktree hook usage

## Tests
- env PKG_CONFIG_PATH=/opt/homebrew/opt/libgit2/lib/pkgconfig rtk cargo test config::tests::config_path -- --nocapture
- env PKG_CONFIG_PATH=/opt/homebrew/opt/libgit2/lib/pkgconfig rtk make test

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,8 @@ st config --set-ai         # pick AI agent + model
 st config --reset-ai       # clear saved AI pairing and re-prompt
 ```
 
-Config lives at `~/.config/stax/config.toml`:
+Config lives at `~/.config/stax/config.toml`. A repo-local `.config/stax/config.toml`
+is used when present, and `STAX_CONFIG_DIR` overrides both:
 
 ```toml
 [submit]
--- a/README.md
+++ b/README.md
@@ -310,8 +310,8 @@ st config --set-ai         # pick AI agent + model
 st config --reset-ai       # clear saved AI pairing and re-prompt
 ```
 
-Config lives at `~/.config/stax/config.toml`. A repo-root `stax.toml`
-overrides only the values it sets, and `STAX_CONFIG_DIR` overrides both:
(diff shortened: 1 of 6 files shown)


cesarferreira:
i like the idea but i dont like the path, perhaps a `stax.toml` in the root the git repo could override only the values it sets would be a better approach to this

gabriel-ecegi:
> i like the idea but i dont like the path, perhaps a `stax.toml` in the root the git repo could override only the values it sets would be a better approach to this

Good idea. Implemented.

cesarferreira:
Hey @gabriel-ecegi, nice feature — clean approach overall. A few things before this lands:

**1. `git_root()` spawns a subprocess on every `Config::load()` call.** The project already has `git2` as a dep — suggest using `git2::Repository::discover("."))` instead to avoid the overhead:
```rust
fn git_root() -> Result<Option<PathBuf>> {
    Ok(git2::Repository::discover(".")
        .ok()
        .and_then(|r| r.workdir().map(PathBuf::from)))
}
```
Also `Path::new(&root).to_path_buf()` → `PathBuf::from(root)` while you're in there.

**2. `config_save_writes_global_config_when_repo_stax_toml_exists` doesn't check saved content** — it only asserts the global config file exists, not what's in it. Should assert the actual written TOML.

gabriel-ecegi:
Addressed both, thanks.

- Replaced the `git rev-parse` subprocess with `git2::Repository::discover(".")`.
- Strengthened the save-path regression test so it now reads/parses the written global TOML and asserts the saved value.
- Added a nested-directory regression test to cover discovering repo-root `stax.toml` from below the repository root.

Local targeted checks pass:
`cargo test config::tests`

PR #419, opened 2026-05-20
fix: avoid duplicate fetch during update

cesarferreira:
## Motivation

`st update` fetched from `origin` during sync, then immediately fetched again during submit. That made update slower and noisy even though the submit phase can reuse refs fetched by update.

- Fetch current-stack remote refs during the update sync fetch, alongside trunk.
- Call submit from update with an internal prefetched mode so normal `st submit` still fetches as before.
- Skip unpublished branch names in the combined fetch so new branches still push cleanly.
- Add integration coverage for existing remote branches and unpublished branches, both asserting update fetches once.

## Validation

- `cargo test --lib`
- `cargo test --test integration_tests test_update -- --nocapture`
- `cargo test --test integration_tests test_submit -- --nocapture`
- `make test`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -94,6 +94,7 @@ impl From<SubmitOptions> for commands::submit::SubmitOptions {
             publish: submit.publish,
             no_pr: submit.no_pr,
             no_fetch: submit.no_fetch,
+            prefetched: false,
             no_verify: submit.no_verify,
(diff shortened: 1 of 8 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #419** 👈

PR #421, opened 2026-05-20
feat(config): add submit.single_stack option

### LINKED ISSUE #240: Config for creating stack comment in PR only if more than one PR in stack
@cesarferreira just picked up this project today and it pretty much check all of the boxes of what I was looking for! Great job working on it so far!

Git Town has a config that tells it to only create the stacked PRs comment/body if there are multiple PRs in the stacks. This avoids a comment on the PR in cases there really isn't much of a stack.

Currently, stax creates a comment like:

> ## Stack Links
> 
> This PR is part of a stacked series:
> 
> * `main`
>   * **PR #1462** 👈
>
>
>This comment was autogenerated by [stax](https://github.com/cesarferreira/stax)

Which kind of doesn't make a lot of sense. stax should add the comment in case other PRs are created stacked to this one.

I'd be happy to contribute with this feature if you feel like it's something worth pursuing.

theoribeiro:
## Summary

Adds a new `submit.single_stack` config option that controls whether stax syncs stack-link comments/body blocks while a stack contains only one PR.

Closes #240 

```toml
[submit]
stack_links = "comment"   # existing — where stack links go
single_stack = "on"       # new — "on" (default) | "off"
```

- **`on` (default):** existing behavior — stack links are synced per `stack_links` on every submit, regardless of stack size.
- **`off`:** stax skips link sync (and cleans up any stale links left over from a previous `on` setting) while the stack has ≤ 1 PRs. As soon as a second PR is submitted on the same stack, links populate on **every** PR — including the one that was previously solo — automatically.

## How it works

In `src/commands/submit.rs` the sync loop now computes an `effective_stack_links_mode`. When `single_stack == Off` and the current stack contains at most one PR, the effective mode is forced to `Off` for that pass — which routes through the existing cleanup branch (delete stack comment, strip body links). When the stack reaches 2+ PRs the override doesn't apply, and because the sync loop already iterates the full current stack, `update_stack_comment` falls back to `create_stack_comment` for any prior PR that has no comment yet. No new code paths — just a gate.

## Changes

- **Config**: new `SingleStackMode` enum (`On` default, `Off`) and `submit.single_stack` field, mirroring the existing `StackLinksMode` shape.
- **Submit**: effective-mode override in the stack-link sync loop in `commands/submit.rs`.
- **Tests**: unit tests for config parsing (defaults, round-trip, rejects unknown), plus three integration tests:
  - `single_stack=off` + solo PR → existing stale comment deleted, body links stripped.
  - `single_stack=off` + 2-PR stack → both PRs get stack-link comments populated (override does not apply).
  - `single_stack=on` (default) + solo PR → comment synced as before (regression guard).
- **Docs**: `docs/configuration/index.md`, `README.md`, and `skills.md` updated to document the new option.

## Test plan

- [x] `cargo test --lib config::` — all 59 config unit tests pass
- [x] `cargo test --test integration_tests forge_mock_tests::test_submit_single_stack` — 3 new integration tests pass
- [x] `make test` — full Docker suite, **1669 tests pass, 0 failed, 0 skipped**
- [x] Dogfooded: submitted this PR via local `stax submit` with `single_stack = "off"` set; verified the resulting PR has no stack-link comment and no body stack-links block (correct behavior for a solo PR).

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -26,6 +26,7 @@ Main config path: `~/.config/stax/config.toml`.
 
 [submit]
 # stack_links = "comment" # "comment" | "body" | "both" | "off"
+# single_stack = "on"     # "on" | "off" — when "off", skip stack-link sync while the stack has only one PR
 
 [ci]
 # alert = false
@@ -155,9 +156,12 @@ Where `st submit` writes the stack graph for a PR:
 ```toml
 [submit]
 stack_links = "body"   # "comment" | "body" | "both" | "off"
+single_stack = "on"    # "on" | "off"
 ```
 
 When body output is enabled, stax appends a managed block to the bottom of the PR body and only rewrites that managed block on future submits.
+
+`single_stack` controls whether stack links are written when the stack contains only one PR. With the default `"on"`, links are always synced per `stack_links`. With `"off"`, stax skips link sync — and removes any stale links left over from a previous `"on"` setting — while the stack has a single PR. As soon as a second PR is submitted on the same stack, links populate on every PR (including the original) automatically.
 
 ## Forge type override
 
--- a/README.md
+++ b/README.md
@@ -322,6 +322,7 @@ Config lives at `~/.config/stax/config.toml`:
 ```toml
 [submit]
 stack_links = "body"   # "comment" | "body" | "both" | "off"
+single_stack = "on"    # "on" | "off" — when "off", skip stack-link sync while only one PR exists
 ```
 
 → [Full config reference](docs/configuration/index.md)
--- a/skills.md
+++ b/skills.md
@@ -200,6 +200,7 @@ stax submit --rerequest-review     # Re-request existing reviewers on update
 # ~/.config/stax/config.toml
 [submit]
 stack_links = "body"               # "comment" | "body" | "both" | "off"
+single_stack = "on"                # "on" | "off" — when "off", skip stack-link sync while only one PR exists; populates on all PRs as soon as the stack reaches 2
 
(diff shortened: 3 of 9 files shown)


docs/configuration/index.md
```diff
@@ -155,10 +156,13 @@ Where `st submit` writes the stack graph for a PR:
 ```toml
 [submit]
 stack_links = "body"   # "comment" | "body" | "both" | "off"
+single_stack = "on"    # "on" | "off"
```
cesarferreira, line 159: can you point out which one is the default in the docs?

cesarferreira:
Thanks for the PR, can you fix the tests?

docs/configuration/index.md
theoribeiro: I wrote it on line 164, want me to make it more clear somewhere else?

> `single_stack` controls whether stack links are written when the stack contains only one PR. With the default `"on"`, links are always synced per `stack_links`. With `"off"`, stax skips link sync — and removes any stale links left over from a previous `"on"` setting — while the stack has a single PR. As soon as a second PR is submitted on the same stack, links populate on every PR (including the original) automatically.

theoribeiro:
> Thanks for the PR, can you fix the tests?

This seems to be a flaky test. It's passing on my computer. I also don't think I touched this test but I'd gladly fix it. 

Any idea why?

theoribeiro:
Ok, tried fixing it by creating empty commits for the test that was failing (that previously created empty files), assuming some of the git commands aren't running properly in the CI. Also added an assertion function to show us if any git command is not working. Hopefully this will help diagnose the issue

PR #434, opened 2026-05-26
Always restack branches when moving to a new parent

### LINKED ISSUE #433: stax move without --restack writes metadata that makes subsequent restack skip the branch
## Bug Description

`stax upstack onto` (and its `stax mv` alias) writes the new parent's revision into the branch metadata **before** the git ref is actually rebased. When `--restack` is not passed, this makes any subsequent standalone `stax restack` or `stax upstack restack` believe the branch is already correctly positioned â€” it skips the branch and the git history is never fixed.

## Steps to Reproduce

1. `git checkout main`
2. `stax create a` â†’ make a commit on `a`
3. `stax create b` (while on `a`) â†’ `b` is created at `a`'s HEAD (same commit)
4. `stax mv b main` (without `--restack`)
5. `stax restack`
6. `git log b` â€” still shows `a`'s commits

## Root Cause

In `src/commands/upstack/onto.rs`, lines 80â€“92:

```rust
let merge_base = repo
    .merge_base(&new_parent, &current)
    .unwrap_or_else(|_| parent_rev.clone());

let updated = BranchMetadata {
    parent_branch_name: new_parent.clone(),
    parent_branch_revision: merge_base.clone(),
    ..old_meta
};

if !restack {
    updated.write(repo.inner(), &current)?;
}
```

After the move, `b`'s metadata says `parent_branch_name=main`, `parent_branch_revision=<main's tip>`. When `needs_restack()` runs later (during `stax restack`), it compares main's current tip against the stored revision â€” they match, so it returns `false` and the branch is skipped. The git ref was never moved.

## Impact

The moved branch retains all commits from its old parent chain. Stack diff output shows commits that belong to a different branch. Users who `move` and then `restack` (as the CLI suggests after a move) get a false sense of correctness.

## Suggested Fixes

Options (from least to most aggressive):

1. **Don't write updated metadata in the no-`--restack` path** â€” leave old parent metadata intact so `needs_restack` correctly detects the drift and restack will rebase it. The tradeoff is that `stax status` will show the branch as needing restack until the user runs restack.

2. **Always restack during move** â€” remove the `--restack` flag and always perform the rebase. This is what most users expect from "move a branch to a new parent."

3. **Fast-forward when the branch has no unique commits** â€” when `merge_base(new_parent, branch)` equals `new_parent`'s tip (i.e., the branch has zero unique commits vs the new parent), do a `git branch -f <branch> <new_parent>` instead of a full rebase, even without `--restack`.

## Verification

Confirmed with direct git test: `git rebase --onto main a b` correctly moves `b` to main's tip (because `a..b` is empty). The upstream calculation in `resolve_rebase_upstream()` correctly identifies `a`'s tip as the fork-point. The fix is purely in the non-`--restack` metadata write path.

## Environment

- `stax` version: 0.80.0
- Branch: main

cesarferreira:
## Motivation

Fixes #433

Moving a branch to a new parent should update both the branch metadata and the underlying git history immediately. Before this change, `stax mv` and `stax upstack onto` could leave the branch pointing at the old ancestry until a separate restack, which produced incorrect history and confusing status output.

- `stax mv` and `stax upstack onto` now always restack after reparenting, so the branch tip and its descendants are rebased onto the new parent in one step.
- The `--restack` flag is kept for backward compatibility but is hidden from help output and no longer changes behavior.
- Added regression tests covering no-op moves, unique commit rebases, subtree restacking, metadata updates, and backward-compatible flag handling.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -836,6 +836,9 @@ enum Commands {
         /// Accepted for backward compatibility; restack now always runs
         #[arg(long, hide = true)]
         restack: bool,
+        /// Stash uncommitted changes before rebasing and restore them after
+        #[arg(long)]
+        auto_stash_pop: bool,
     },
 
     /// Interactively reorder branches within a stack
@@ -1442,6 +1445,9 @@ enum UpstackCommands {
         /// Accepted for backward compatibility; restack now always runs
         #[arg(long, hide = true)]
         restack: bool,
+        /// Stash uncommitted changes before rebasing and restore them after
+        #[arg(long)]
+        auto_stash_pop: bool,
     },
 
     /// Submit current branch and descendants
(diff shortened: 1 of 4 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #434** 👈

PR #435, opened 2026-05-27
recover existing PR on duplicate submit create

cesarferreira:
## Motivation

`stax submit` can successfully create a PR on GitHub but still end by surfacing a duplicate PR validation error if the create result is not reconciled and a follow-up create attempt sees the same head branch. In that state the PR is real, but local stax metadata is left stale and the command appears to have failed.

- Treat GitHub duplicate-PR create responses for the same head branch as an idempotent submit race.
- On that duplicate response, look up the open PR by head branch, persist the discovered PR metadata, and continue through the normal stack-link sync path.
- Keep ordinary successful creates and non-duplicate create failures on their existing behavior.
- Add a mock GitHub regression test that reproduces the race: planning sees no PR, create returns `A pull request already exists`, the follow-up head lookup finds the PR, and metadata records that PR number.
- Add a repo lesson documenting that submit/create flows should be idempotent around duplicate PR responses.

## Why this should not destroy unrelated behavior

- The new branch only runs when `create_pr` returns an error whose message contains both `A pull request already exists` and `PullRequest`.
- The normal `Ok(pr)` create path is unchanged and still marks the PR as newly created for stack-comment behavior.
- Non-duplicate create errors still return through the same contextual failure path.
- The recovered PR then uses the existing metadata write and stack-link sync code, rather than a separate mutation path.

## Verification

- `make test` (Docker): 1693 tests run, 1693 passed, 0 skipped
- `cargo test --test integration_tests forge_mock_tests::test_submit_adopts_existing_pr_when_create_reports_duplicate_head -- --nocapture`
- `cargo test --test integration_tests forge_mock_tests::test_submit -- --nocapture` (24 passed)
- `cargo test submit:: --lib` (27 passed)
- `cargo fmt --check`
- `git diff --check main...HEAD`

Note: the first Docker run hit a stale/corrupt `.docker-cache/target` compile artifact (`can't find crate for serde`). Moving that cache aside and rerunning `make test` produced the passing full-suite result above.


cesarferreira:
## Stack Links

* `main`
  * **PR #435** 👈

PR #436, opened 2026-05-27
Sync refreshes tracked PR metadata before completing

cesarferreira:
## Summary
This change fixes `sync` so PR metadata refresh is bounded, visible in progress output, and completed before the final `Sync complete!` footer is printed.

## What Changed
- Moved PR metadata refresh into the main timed sync flow so it contributes to overall command timing.
- Added a dedicated `Refresh PR metadata` progress step for human-facing output.
- Parallelized live PR lookups with a bounded concurrency limit of 8 to avoid sequential refresh latency.
- Switched refresh logic to target only locally tracked PR branches instead of scanning repo-wide open PR state.
- Extracted branch/CiCache reconciliation into `apply_live_pr_state` to keep the refresh path clearer and easier to maintain.
- Added regression coverage verifying:
  - the refresh step is visible before `Sync complete!`
  - tracked PRs are fetched directly by PR number
  - the sync refresh does not scan the full open-PR list
- Documented the durable lessons in `learnings.md`.

## Why
Previously, sync could hide a best-effort metadata refresh after the completion footer and incur unnecessary latency by working from repo-wide PR state. The new flow makes the work visible, finishes it before the command reports completion, and scales with the set of locally tracked branches.

## Verification
- Added an integration test covering the refresh timing and request shape.
- Existing sync behavior remains covered by the integration suite around branch state, PR state, and metadata reconciliation.

## Notes
- PR metadata refresh remains best-effort: failures are skipped rather than failing the whole sync.
- The refresh still updates branch metadata and CiCache when live PR data is available.

### DIFF
--- a/learnings.md
+++ b/learnings.md
@@ -16,6 +16,8 @@
 - When adding or changing CLI commands/flags, update both `README.md` and `docs/` command references in the same change and verify parity against `stax --help` before marking docs complete.
 - AI standup prompts that ask for `Today` must include concrete in-flight signals (recent branch work, opened PRs, or Jira tickets with PRs) and tell the model to carry unfinished work forward; otherwise the model fills missing context with generic review/cleanup/backlog bullets.
 - Long-running reporting or AI commands must show progress for human-facing output while keeping `--json` and plain-text modes free of progress noise; use stderr for progress when stdout may be copied, piped, or parsed.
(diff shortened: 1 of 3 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #436** 👈

PR #465, opened 2026-06-04
fix: replace stale `stax shell-setup --install` hints with `stax setup`

### LINKED ISSUE #299: fix: replace stale  hints with a real setup command
Several shell-integration hints still recommend `stax shell-setup --install`, but `setup` no longer has an `--install` flag.

Evidence in the current codebase:
- `src/commands/shell_setup.rs:475`
- `src/commands/worktree/go.rs:119`
- `src/commands/worktree/create.rs:75`
- `src/commands/worktree/create.rs:115`
- `src/commands/worktree/create.rs:208`
- matching assertions in `tests/worktree_cli_tests.rs:662` and `:689`

Why this matters:
- users who follow the hint hit an invalid command right after `st` tells them how to enable automatic `cd`
- it is especially visible in worktree flows, where shell integration is most relevant
- the CLI already documents `st setup` / `st shell-setup` as the onboarding entry point, so these strings are now inconsistent with the rest of the docs

Proposed improvement:
- replace the stale hint with a real command, likely `st setup` or `st shell-setup`
- update the related worktree tests to assert the new guidance
- optionally add a small regression test so future onboarding renames keep these helper messages in sync

Expected benefit:
- less onboarding friction
- fewer “command/flag does not exist” dead ends
- consistent shell-integration guidance across docs, prompts, and worktree UX

cesarferreira:
## Summary

- **Fixes #299**: replace 5 stale `stax shell-setup --install` hints (in `shell_setup.rs`, `worktree/go.rs`, `worktree/create.rs` ×3) with `stax setup`. The `--install` flag doesn't exist — only `--install-skills` does — so users who followed the hint hit `error: unexpected argument '--install' found` right after stax told them how to enable transparent `cd`. Updates the 2 worktree-CLI test assertions to match.
- **Docker test env**: bumps `make test-docker` from `rust:1.93` → `rust:1.95` to match the locally-installed toolchain, and adds a minimal `Dockerfile` (rust:1.95 + cargo-nextest) that captures the same env for future CI reuse. The Makefile still inlines the image for now so day-to-day usage is unchanged.
- **AGENTS.md**: notes that the Docker daemon must be running before `make test`, so agents stop silently falling back to native runs when the socket isn't available.

## Test plan

- [x] `cargo run -- setup` confirms `stax setup` is the working canonical command (the hint now points somewhere real).
- [x] `cargo test --test worktree_cli_tests suggests_install` — both hint-assertion tests pass.
- [x] `make test` on `rust:1.95` — full suite, 1767 tests, all green.
- [x] `grep -rn "shell-setup --install" src/ tests/ docs/ README.md` returns nothing.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,6 +3,7 @@
 ## Test Command Policy
 
 - **AI agents:** for full-suite validation always run `make test`. On macOS this routes through Docker, which is the only sane way to run the entire integration suite — `cargo test` natively will be slow, flaky, and may exhaust file handles. Default to `make test` and only fall back to native runners when explicitly told to.
+- **Start Docker before running `make test`.** On macOS the Docker daemon is not always running; if `make test` fails with `failed to connect to the docker API at unix:///.../docker.sock`, ask the user to launch Docker Desktop (or run `open -a Docker`) and retry — do not silently fall back to `make test-native`.
 - Do not run the full suite via `cargo test` in this repo.
 - For full-suite validation, always use `make test`.
 - On macOS, `make test` intentionally routes to the Docker fast path.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM rust:1.95
+
+RUN cargo install cargo-nextest --locked
+
+WORKDIR /work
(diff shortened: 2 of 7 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #465** 👈

PR #376, opened 2026-05-13
Show pull request changes-requested status correctly

cesarferreira:
## Motivation

Pull request review status could be reported incorrectly when a stale review still contained `CHANGES_REQUESTED`, even after the overall PR review state had changed.

## Description of changes / notes for reviewers

- Use GitHub's `reviewDecision` to determine whether changes are currently requested.
- Keep approval counting unchanged, while avoiding false positives from older review events.

PR #447, opened 2026-06-02
feat: add PR readiness view

cesarferreira:
## Summary

- docs: specify PR readiness table

PR #469, opened 2026-06-04
fix(github): resolved changes-requested reviews no longer block PRs (re-fix #376)

theoribeiro:
## Summary

- **Re-fixes #376, regressed by #447**: `get_pr_merge_status` was rewritten to GraphQL in #447 and reintroduced the stale-review scan — `changes_requested` was OR'ing `reviewDecision` with an `any()` scan over the full historical `reviews(last: 100)` list. A reviewer who requested changes and later approved kept the PR flagged as blocked in `stax ready` and the merge commands. `changes_requested` now derives solely from GitHub's `reviewDecision` (which already applies per-reviewer latest-wins), matching `get_pr_reviews`. A code comment documents the invariant so it survives the next rewrite.
- **Fixes approvals over-count in the same block**: `approvals` counted historical `APPROVED` events, so one reviewer re-approving after a push counted twice, and an approval superseded by `CHANGES_REQUESTED` still counted. New shared `count_effective_approvals` helper applies per-reviewer latest-wins (only `APPROVED`/`CHANGES_REQUESTED`/`DISMISSED` change standing; `COMMENTED`/`PENDING` can't clear an approval). Both GraphQL queries now fetch `author { login }` (nullable for ghost users).

## Test plan

- [x] New regression test `test_get_pr_merge_status_ignores_stale_changes_requested_review` — verified to fail on the pre-fix code (this is the test #447 lacked)
- [x] 4 new approval-counting tests: repeat approvals → 1, superseded approval → 0, comment doesn't clear approval → 1, distinct reviewers → 2
- [x] Existing fixtures updated with review authors; `cargo test` fully green (1,500+ tests)

### DIFF
--- a/src/commands/sweep.rs
+++ b/src/commands/sweep.rs
@@ -35,11 +35,7 @@ pub fn run(
     // --- Classify all local branches ---
 
     // 1. Merged (ancestor or squash-merged)
-    let merged_infos = find_merged_branches_all(
-        &workdir,
-        &trunk,
-        Some(remote_trunk_ref.as_str()),
-    )?;
+    let merged_infos = find_merged_branches_all(&workdir, &trunk, Some(remote_trunk_ref.as_str()))?;
     let merged_set: HashSet<String> = merged_infos.iter().map(|m| m.branch.clone()).collect();
 
     // 2. Upstream-gone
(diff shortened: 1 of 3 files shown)


cesarferreira:
Thanks for the careful re-fix, @theoribeiro! Clean diagnosis of the #447 regression, and the shared `count_effective_approvals` helper plus the invariant comment should keep this from regressing again. Test coverage is exactly what was missing the first time around. Merging.

PR #476, opened 2026-06-08
[codex] fix sweep upstream-gone safety

### LINKED ISSUE #475: stax sweep marks upstream-gone branches with local-only commits as safe to delete
Category: bug

Found by the nightly Hermes self-improvement scout.

## Evidence

`stax sweep` currently treats every branch whose upstream is `[gone]` as safe-to-delete:

- `src/engine/branch_detect.rs:89-115` classifies upstream-gone branches only by `%(upstream:track)` containing `[gone]`.
- `src/commands/sweep.rs:45-51` places those branches in the `upstream-gone` bucket unless already merged/current/trunk.
- `src/commands/sweep.rs:221-228` includes all upstream-gone branches in the deletion set for `stax sweep --delete`.
- `src/commands/sweep.rs:488-490` deletes via Git's forced branch deletion path, so local-only commits are not protected by Git's normal merged check.

I reproduced a branch whose remote was deleted but which still had two local commits not on trunk. `stax sweep` still called it safe to delete:

```text
Branch sweep — stale threshold: 30 days

  upstream-gone  (1) — remote deleted, safe to delete
    ⚑ gone-with-local-work

Tip: run stax sweep --delete to delete 1 merged/gone branch
ahead-behind main...gone-with-local-work: 0	2
```

The `git rev-list --left-right --count main...gone-with-local-work` result shows the branch is 2 commits ahead of `main`, despite being classified as safe.

## Why it matters

A user can reasonably have local follow-up commits after pushing a branch. If the remote branch/PR is deleted or renamed, `stax sweep --delete` with auto-confirm can remove that local-only work while presenting it as a safe cleanup. The command is marketed as branch housekeeping, so the safety bar should be high.

## Suggested fix / test plan

- Split upstream-gone into at least two cases:
  - `upstream-gone + merged/no unique commits` => safe deletion candidate.
  - `upstream-gone + local-only commits` => active/warning bucket; do not include in `--delete` unless a stronger explicit flag exists.
- Before adding an upstream-gone branch to the default deletion set, verify it is merged into trunk/remote trunk or has no unique commits relative to its configured merge base/trunk.
- Add an integration test that:
  1. creates and pushes a branch,
  2. adds an additional local-only commit,
  3. removes the remote ref and fetches/prunes,
  4. asserts `stax sweep` does not label it as safe/delete it by default.

cesarferreira:
## Summary
- Protect upstream-gone branches from `stax sweep --delete` when they have commits not reachable from local or remote trunk
- Keep protected upstream-gone branches classified as active so they are not deleted by default or via `--include-stale`
- Update sweep docs and agent-facing command notes to describe the narrowed deletion set

## Root cause
`sweep` treated every `[gone]` upstream as safe to delete and later used `git branch -D`, which could remove local-only work after a remote branch disappeared.

Fixes #475.

## Verification
- `cargo test --test sweep_tests -- --nocapture`
- `make test` (Docker nextest: 1793 tests run, 1793 passed, 0 skipped)

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ st config --set-ai
 | `st ci -w --alert` | Watch CI until all checks finish, then play success/error sounds |
 | `st ci -w --strict` | Watch CI but exit as soon as any check fails |
 | `st rs` / `st rs --restack` | Sync trunk, clean merged branches, optionally rebase |
-| `st sweep` | Classify all local branches (merged/gone/stale/active); `--delete` to remove safe-to-delete ones |
+| `st sweep` | Classify all local branches (merged/gone/stale/active); `--delete` removes merged branches and upstream-gone branches with no unique work |
 | `st update` | Sync trunk without merged-branch cleanup, restack current stack, then push/update PRs |
 | `st update --force --yes --no-prompt` | Run update without sync or submit prompts |
 | `st update --verbose` | Include detailed sync/restack/submit timing |
--- a/docs/commands/core.md
+++ b/docs/commands/core.md
@@ -56,7 +56,7 @@ When `-m` or `--ai` derives a branch name that already exists, Stax stops instea
 | Command | What it does |
 |---|---|
 | `st sweep` | Classify all local branches: merged, upstream-gone, stale, active (read-only) |
-| `st sweep --delete` | Delete all merged and upstream-gone branches after confirmation |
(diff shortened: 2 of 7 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #476** 👈

PR #486, opened 2026-06-10
update cargo lock

cesarferreira:
## Motivation

Update the GitHub client to work with `octocrab` 0.53.1 and bring the dependency stack forward. This also addresses API shape changes in newer octocrab responses, especially around optional PR fields and GraphQL error handling, so PR operations keep behaving reliably after the upgrade.

- Bumped `octocrab` from `0.49` to `0.53.1` and refreshed `Cargo.lock` to the newer transitive dependency set.
- Adapted PR parsing to the newer octocrab model types where `number`, `head`, and `base` can be optional, skipping incomplete PR records and surfacing a clear error when expected fields are missing.
- Consolidated repeated PR mapping logic into helper functions and switched GraphQL calls to a shared helper that unwraps octocrab GraphQL errors into consistent `anyhow` failures.
- Updated PR lookup, listing, creation, merge-queue, review, and node-id flows to use the new helpers without changing the public behavior of the GitHub client.
- Added coverage for the new missing-head failure path in `get_pr_with_head`, alongside small test cleanups from the refactor.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,6 +89,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "approx"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #486** 👈

PR #488, opened 2026-06-10
[codex] Disable Windows release builds

cesarferreira:
## Summary

- Disable the Windows target in the release workflow matrix while leaving the commented block in place for easy re-enable.
- Update README and docs to say Windows prebuilt binaries are not published.
- Add Windows source-install instructions using `cargo install stax --locked --features vendored-openssl` or a local checkout.

## Why

The Windows release job is currently the slowest release build, and there may not be enough Windows usage to justify publishing prebuilt Windows artifacts. Windows users can still install stax by building it locally.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yml"); puts "release.yml parses"'`
- `git diff --check`
- `rg -n "download .*windows|Windows.*download|stax-x86_64-pc-windows-msvc\.zip|ships prebuilt Windows" README.md docs .github/workflows skills.md` returned no matches

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,8 +29,11 @@ jobs:
             os: ubuntu-latest
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-24.04-arm
-          - target: x86_64-pc-windows-msvc
-            os: windows-latest
+          # Windows builds are intentionally disabled to keep releases fast.
+          # Windows users can build from source with:
+          # cargo install stax --locked --features vendored-openssl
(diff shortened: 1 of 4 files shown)


geoHeil:
@cesarferreira I actually would love to keep having windows builds - no chance to keep it enabled?

cesarferreira:
the windows builds take ~3times longer than linux and macos and I didnt know if people were using windows with stax so I removed it but if you use it i can restore it

geoHeil:
thank you!

and do you see any optimization potential for the windows builds?

PR #489, opened 2026-06-10
perf(build): cut test link time with line-tables-only debuginfo

cesarferreira:
## Motivation

Reduce local and test build time by cutting the cost of linking the many integration-test binaries. Full debuginfo was making relinks unnecessarily slow, especially on macOS.

## Description of changes / notes for reviewers

- Set `[profile.dev]` and `[profile.test]` to use `debug = "line-tables-only"` so panic backtraces keep file and line information without full debuginfo overhead.
- Set `split-debuginfo = "unpacked"` to skip the extra dsymutil packaging step on macOS and further reduce link time.

PR #490, opened 2026-06-10
perf(test): consolidate integration tests into a single binary

cesarferreira:
## Motivation

Reduce integration test build time by avoiding the cost of compiling and linking ~50 separate test binaries. This also standardizes how the suite is run locally now that all integration tests live under one consolidated target.

- Disable Cargo’s integration-test auto-discovery and register a single `all_tests` target in `Cargo.toml`, so the full suite compiles into one binary instead of one binary per file.
- Add `tests/all_tests.rs` as the central integration-test entrypoint and include each `tests/*_tests.rs` module there, with shared helpers exposed once through `crate::common`.
- Update every integration test file to import `common` from the consolidated test crate instead of declaring its own local `mod common;`.
- Refresh the repo guidance in `AGENTS.md` to reflect the new test layout and the recommended `cargo nextest run <pattern>` workflow for targeted runs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,7 +11,8 @@
   - `make test-native`
   - `make test-local-ramdisk`
   - `make test-local-fast`
-- Targeted single-test or single-crate runs via `cargo test --test <name>` / `cargo nextest run <pattern>` are fine and encouraged for tight feedback loops; switch to `make test` once changes are ready for verification.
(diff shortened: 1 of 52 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #489**
    * **PR #490** 👈
      * **PR #491**

PR #518, opened 2026-06-22
fix: show git push hook output on submit failure

### LINKED ISSUE #517: submit hides local pre-push hook output when git push fails
## Summary

`stax submit` can collapse a local `git push` failure into a generic message such as `Failed to push branches <branch>: error: failed to push some refs`. When the real blocker is a local pre-push hook, the actionable hook output may be printed on stdout and is not visible in the final Stax error.

## Sanitized Example

A branch submit runs `git push --porcelain --force-with-lease -u <remote> <branch>`. A local pre-push hook runs an audit tool and exits non-zero. The audit tool prints a vulnerability report to stdout. Stax reports only the generic push failure, while running `git push` manually shows the audit report and makes the cause obvious.

## Expected Behavior

When the underlying `git push` command exits non-zero, Stax should preserve useful output from both stdout and stderr in the final submit error. This is especially important for local hooks, because hook wrappers and audit tools do not consistently use stderr for human-facing failure details.

## Notes

This is not asking Stax to bypass hooks. The goal is to show the actual reason the push was rejected so users do not have to rerun `git push` manually to diagnose submit failures.

geoHeil:
## Summary

- preserve both stdout and stderr from failed `git push` calls during submit
- keep rejected-branch detection while appending the underlying push/hook output
- cover stdout-only and mixed-stream hook diagnostics in tests
- update the worktree random-lane test for the current no-file-count creation output

Fixes #517.

## Tests

- `cargo nextest run submit_no_verify_tests:: push_failure_details`
- `cargo nextest run worktree_cli_tests::wt_create_without_name_creates_random_lane`
- `git diff --check`
- `make test` (Docker): 1684 tests passed

## Docs

No README/docs/skills update: this improves submit failure diagnostics and repairs a stale test assertion; it does not add or change commands, flags, workflows, concepts, or defaults.

### DIFF
--- a/src/commands/submit.rs
+++ b/src/commands/submit.rs
@@ -1848,27 +1848,23 @@ fn push_branches(
         let stdout = String::from_utf8_lossy(&output.stdout);
         let stderr = String::from_utf8_lossy(&output.stderr);
         let rejected = rejected_push_branches(&stdout, branches);
-        let details = stderr.trim();
+        let details = push_failure_details(&stdout, &stderr);
+        let branch_list = branches.join(", ");
         if !rejected.is_empty() {
-            anyhow::bail!(
-                "Failed to push branches {}: rejected {}{}",
-                branches.join(", "),
-                rejected.join(", "),
-                if details.is_empty() {
-                    String::new()
-                } else {
-                    format!(" ({})", details)
-                }
+            let mut message = format!(
+                "Failed to push branches {branch_list}: rejected {}",
+                rejected.join(", ")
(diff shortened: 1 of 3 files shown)


cesarferreira:
@copilot resolve the merge conflicts in this pull request

cesarferreira:
Manually merged after resolving the conflict with #513 (which had replaced the same `"and copied"` assertion with `"Branched" && "from main"`). Changes landed in a9ea2db.

PR #551, opened 2026-07-01
Switch changelog and release notes to git-cliff

cesarferreira:
Replaces the custom `scripts/prepare_release.py` changelog generator with [git-cliff](https://git-cliff.org/), and makes it the single source of truth for both the committed `CHANGELOG.md` and the GitHub Release notes.

## Why

The Python script maintained a bespoke Keep-a-Changelog format (category mapping, compare-link footers, `Unreleased` juggling) that had to be hand-maintained and had drifted (e.g. duplicated 0.90/0.91 entries). git-cliff is a well-supported, config-driven generator that integrates cleanly with `cargo-release`.

## Changes

**Changelog generation**
- Add `cliff.toml` — grouped conventional-commit style, GitHub PR links, and parsers that skip the automated release commits, merge commits, and dependabot bumps. `filter_unconventional = false` keeps this repo's many non-conventional squash-merge subjects (e.g. `Fix X (#123)`) in an `Other` group instead of silently dropping them.
- Regenerate `CHANGELOG.md` fully from git history in the new style.
- Point `cargo-release`'s pre-release hook (`release.toml`) at `git-cliff -o CHANGELOG.md --tag v{{version}}`.
- Remove `scripts/prepare_release.py` and its now-obsolete integration tests (`tests/release_prep_tests.rs`).

**Release workflow (`.github/workflows/release.yml`)**
- Check out the `release` job with `fetch-depth: 0` (git-cliff needs full history + tags).
- Generate the release body with `orhun/git-cliff-action@v4` (`--latest --strip header`).
- Feed that into `softprops/action-gh-release` as `body:`, replacing `generate_release_notes: true`.

**Tooling — no OS-level dependency**
- CI already brings its own git-cliff via `orhun/git-cliff-action` (no PATH dependency).
- For local `make release`, the target now depends on `ensure-git-cliff`, which installs a pinned git-cliff via `cargo install` when it is missing — so contributors don't need to install it through brew/apt.

**Docs**
- Update `docs/workflows/releasing.md`, `README.md`, and `skills.md` to describe the git-cliff flow.

## Testing

- Regenerated `CHANGELOG.md` and verified: no leaked commit bodies, no release/merge commits, PR references resolve to repo links.
- Simulated the exact `cargo-release` hook (`git-cliff -o CHANGELOG.md --tag v0.92.0`) and confirmed it produces a correct `[0.92.0]` section grouping unreleased commits.
- `cargo test --no-run --test all_tests` compiles clean after removing the obsolete tests.
- Did **not** run a real `cargo release` or push a tag, so the `release.yml` git-cliff step will be exercised for the first time on the next release.

## Behavior changes to note

- git-cliff does not fail-early on an empty commit range the way the old script did (`No commits found since last tag`); it renders an empty section instead.


cesarferreira:
## Stack Links

* `main`
  * **PR #551** 👈

PR #560, opened 2026-07-02
feat: add config edit option for custom AI models

### LINKED ISSUE #559: [FEAT] Allow usage of other models in the harnesses
## Problem

Model selection for AI features is currently limited

Image

Shows OpenCode allowing only selection of default models.

OpenCode actually has other models available

Image

## Solution (?)

It is possible to configure this in the ` ~/.config/stax/config.toml` (I think), but having an `(edit config file)` entry in the menu would make this more obvious

cesarferreira:
## Motivation

The interactive model picker only exposes known model IDs, so custom providers and newly released models require a clear path to edit configuration without losing the selected feature scope.

- add an “edit config” choice after known models, including the empty-model case
- initialize and open the user config with the configured editor, then reload the edited values
- resolve exact feature and global model scopes without accidentally inheriting the global model for lanes
- preserve unrelated feature settings when persisting the selected custom model
- provide contextual errors for missing editors, editor failures, malformed config, and unreadable config
- document custom model setup for OpenCode, pi, PR generation, configuration, README, and `skills.md`

## Verification

- `cargo check`
- `make lint-fast`
- focused model-picker, editor, reload, and scope tests — 17 passed
- `git diff origin/main --check`
- refreshed branch is a fast-forward from the previous PR head; no force-push

Closes #559


CharlieDigital:
One question here: would it be possible to add a "--test" flag (or is there a good way to do this quickly?  The challenge is getting the configuration correct and the harnesses (like opencode) have their own naming for the models.  I think this is a good addition (I already fixed it for myself :) ), but good for other folks.

cesarferreira:
what do you mean by the `--test` ?
how did you fix it? if it's a good solution for the users you could open a PR ;)

PR #565, opened 2026-07-02
feat: native GitHub Stacked PR integration

cesarferreira:
Clean reimplementation of native GitHub Stacked PR support, replacing #564 (reverted in #566).

Registers stax-managed stacks with GitHub's native Stacked PRs feature via the `github/gh-stack` extension (`gh stack link`). Verified against `github/gh-stack v0.0.7`; built only on commands/behavior that the real extension actually provides.

## What it does

- **`stax submit`** auto-links multi-PR stacks when the extension is present and the repo is eligible (`[submit] native_stack = "auto"`). Feature-disabled repos are cached so future submits stay quiet. `--native-stack` / `--no-native-stack` override per run.
- **`stax stack link` / `stax stack unlink`** for manual control.
- **`stax doctor`** reports extension readiness (installed / missing / outdated); `stax doctor --fix` installs or upgrades it.
- Keeps stax body/comment stack links alongside the native map by default (`stack_links_when_native`).

## Correctness details (the gaps that sank #564)

- **≥2 PRs required.** `gh stack link` needs two or more PRs. Single-PR stacks are skipped on submit and rejected by `stax stack link` with a clear message (no raw `requires at least 2 arg(s)` leak).
- **Outdated extension.** `gh stack link` postdates `gh-stack v0.0.1`; older versions are detected (probing `gh stack --help`) and reported as outdated, with `doctor --fix` running `gh extension upgrade`.
- **No fictional read API.** `gh stack view --json` has no per-PR lookup, so there is no `st get`/`st merge` "native discovery" — those paths from #564 were dropped rather than shipped broken.
- **Honest unlink.** `gh stack unstack` needs a locally-tracked stack, which `gh stack link` doesn't create; `stax stack unlink` explains this instead of erroring cryptically or risking a destructive delete.

## Test plan

- [x] `cargo check --all-targets`
- [x] `cargo nextest run gh_stack_tests::` — 13 pass (fake `gh` shim + wiremock): extension detection (installed/outdated), link arg ordering, feature caching, doctor install/upgrade, single-PR handling, submit auto-link keeping stax links, feature-disabled caching.

## Related

- Reverts/replaces #564
- Revert PR: #566

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ One giant PR is slow to review and risky to merge. A stack of small PRs is the a
 - **Stack, don't wait.** Keep shipping on top of in-review PRs. `st create`, `st ss`, done.
 - **Native-fast.** A single Rust binary that starts in ~25ms. `st ls` benches ~70× faster than Graphite and ~215× faster than Freephite on this repo.
 - **Agent-native.** Run parallel AI agents on isolated branches (`st lane`), auto-resolve rebase conflicts (`st resolve`), and generate branch names, commit messages, and PR details from real diffs.
+- **GitHub-native stacks, automatically.** When your repo has GitHub's native Stacked PRs enabled and `github/gh-stack` is installed, `st ss` registers the stack with GitHub under the hood — zero config, zero extra commands. See [Native GitHub Stacked PRs](#native-github-stacked-prs).
 - **Undo-first.** Every destructive op snapshots state. `st undo` / `st redo` rescue risky rebases instantly.
 - **Batteries-included TUI.** Run bare `st` to browse the stack, inspect diffs, and watch CI hydrate live.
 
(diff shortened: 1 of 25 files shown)


geoHeil:
but this is private preview only right? https://github.github.com/gh-stack/ so is this optional? are you not breaking this for anyone who is not part of the private preview?

cesarferreira:
## Stack Links

* `main`
  * **PR #565** 👈
    * **PR #587**

cesarferreira:
> but this is private preview only right? [github.github.com/gh-stack](https://github.github.com/gh-stack/) so is this optional? are you not breaking this for anyone who is not part of the private preview?

The goal is to have it as an optional under the hood work, if we have the tool installed + the repo supports it it does the extra work, if not it should behave exactly the same.

We have access in my current job and it works nicely there.

PR #616, opened 2026-07-13
docs(gui): design phase 4 distribution

cesarferreira:
## Motivation

The GPUI app can now perform everyday and structural stack work, but it still needs a stable macOS identity and a repeatable release path. This PR turns the developer preview into a separately downloadable desktop artifact without changing the existing CLI packages.

## Description of changes / notes for reviewers

- Finalize the `Stax.app` identity, bundle metadata, version/build-number handling, startup path contract, and single-instance launch behavior.
- Add production app icons and deterministic bundle assembly.
- Build and validate architecture-specific `Stax-aarch64-apple-darwin.zip` and `Stax-x86_64-apple-darwin.zip` artifacts in the release workflow.
- Support unsigned releases by default, optional Developer ID signing, and optional notarization when the complete credential set is configured.
- Validate archive structure, bundle ID, version, executable architecture, executable version output, and an 80 MiB archive-size ceiling before publishing.
- Make visible GPUI actions keyboard-operable and finish the installation, GUI, release, and agent-facing documentation.

The existing CLI archives remain unchanged; the desktop app is an additional GitHub Release artifact. Partial signing or notarization configuration fails closed instead of silently producing a differently trusted build.

## Test plan

- [x] `make lint`
- [x] `make test` — 2,037 tests passed
- [x] `cargo nextest run -p stax-gui --locked`
- [x] `bash scripts/gui-app-tests.sh`
- [x] `bash scripts/gui-release-tests.sh`
- [x] `bash scripts/gui-release-workflow-tests.sh`
- [x] GitHub Actions: formatting/Clippy, unit tests, integration tests, and GPUI/release packaging checks all pass

## Stack

Depends on #615 and is the top of this GUI stack.

PR #615, opened 2026-07-13
docs(gui): design phase 3 parity

cesarferreira:
## Motivation

Phase 2 covers everyday actions, but structural stack changes still required dropping back to the CLI. This PR closes that gap so the GPUI app can safely reshape a stack and recover from those mutations.

## Description of changes / notes for reviewers

- Add shared application operations for branch rename, branch deletion, stack reorder, subtree moves, undo, and redo, with typed receipts and recovery guidance.
- Delegate the corresponding CLI and TUI commands to those shared operations to preserve existing behavior and avoid parallel Git implementations.
- Add GPUI confirmation and progress flows for structural actions, including branch-aware validation and failure reporting.
- Add stack search plus persisted pane-visibility preferences.
- Add native macOS menus, keyboard equivalents, and an About dialog so visible actions are reachable without pointer-only interaction.
- Update the README, command reference, GUI guide, and agent-facing command map for the new structural workflows.

The main review boundary is the application layer: mutation semantics and recovery remain centralized there, while the GPUI layer only drives and renders those operations.

## Test plan

- [x] `make lint` on the submitted stack
- [x] `make test` — 2,037 tests passed on the stack tip
- [x] GitHub Actions: formatting/Clippy, unit tests, integration tests, and GPUI checks all pass
- [x] Structural operation coverage includes success, validation failure, rollback/recovery, and undo/redo paths

## Stack

Depends on #612. Followed by #616.

PR #612, opened 2026-07-12
feat: add GPUI everyday stack operations

cesarferreira:
## Motivation

Phase 1 established the read-only GPUI workspace. This PR makes the desktop app useful for everyday stack work while keeping repository mutations behind the same safety contracts as the CLI and TUI.

## Description of changes / notes for reviewers

- Introduce shared application operations for checkout, branch creation, restack, submit, and pull-request resolution, with typed receipts, errors, progress events, and rollback-aware mutation guards.
- Route the existing CLI and TUI paths through those shared operations so the desktop app does not implement a second set of Git semantics.
- Add GPUI confirmations, text input, progress and result overlays, contextual controls, and keyboard shortcuts for everyday stack actions.
- Add the `st gui [PATH]` launcher plus scripts and documentation for assembling and installing an unsigned developer `Stax.app`.
- Add application-boundary linting and broad integration coverage, including the macOS-only launcher fixture used by the GUI command tests.

Reviewing the shared application contracts first, then the CLI/TUI delegation, and finally the GPUI wiring gives the clearest path through this PR. Release distribution is intentionally deferred to Phase 4.

## Test plan

- [x] `make lint`
- [x] `make test` — 2,037 tests passed
- [x] `cargo nextest run gui_command_tests::` — 6 tests passed
- [x] GitHub Actions: formatting/Clippy, unit tests, integration tests, and GPUI checks all pass

## Stack

Depends on #611. Followed by #615.

PR #611, opened 2026-07-12
feat: add GPUI desktop read-only cockpit

cesarferreira:
## Summary
- extract a typed `stax::application` read layer and migrate TUI detail, diff, and CI hydration onto it
- add the standalone GPUI workspace with recents, a virtualized three-pane cockpit, bounded async hydration, and accessible mouse/keyboard controls
- make shared CI/diff caches transactional and revision-aware while enforcing trusted automatic forge credential routing

## Test plan
- [x] `make test` — 1,938 tests passed
- [x] `make lint`
- [x] `cargo nextest run -p stax-gui --locked` — 94 tests passed
- [x] `cargo check -p stax-gui --locked`
- [x] CI-equivalent GUI Clippy with warnings denied

## Stack
- Stacked on #607, which remains design-only.

## Documentation
README and `skills.md` are unchanged because Phase 1 does not expose a released GUI command or installation path yet; the design and Phase 1 implementation plan document the internal behavior.

## Notes
GPUI 0.2.2 currently brings in `block` 0.1.6, which emits a non-blocking future-incompatibility warning and should be revisited with a future GPUI upgrade.

Made with [Cursor](https://cursor.com)

PR #617, opened 2026-07-13
perf(gui): reuse cached branch diffs instantly

cesarferreira:
## Summary

- keep the 32 most recently viewed branch diffs in GUI session memory so revisiting a branch renders synchronously;
- replace the aggregate diff-cache JSON with one revision-keyed JSON file per diff under `.git/stax/diff-cache/v1`;
- bound persistent storage to 128 entries and 100 MiB with concurrency-safe cleanup and orphan-lock reaping.

## Why

The previous cache stored every patch in one JSON document. In the repository used to reproduce the issue, that file was 20.3 MB with 27 entries and 166,175 patch lines. Selecting any branch had to read and deserialize the entire file before showing one cached patch, taking roughly 260 ms in a command-line parse and making repeated branch selection feel uncached.

## Implementation

- Persistent entries are keyed by the parent, branch, and merge-base object IDs, so unchanged revisions remain safe to reuse across branch renames and app launches.
- Each entry is read, locked, and written independently using the existing atomic persistence and `fs4` locking primitives; no database or new dependency is introduced.
- Cleanup coordinates across processes, ignores foreign files, removes cache-owned orphan locks, and safely degrades malformed or unavailable cache entries to misses without hiding Git errors.
- The GUI uses a snapshot-scoped 32-entry LRU keyed by branch and parent, clears it after repository mutations or snapshot replacement, and retains stale-generation rejection.
- The legacy aggregate file is disposable cache data: it is not migrated, and is removed best-effort after a successful new-format write. The first lookup after upgrading may therefore be cold once.

## Verification

- `cargo fmt --all -- --check`
- `make lint`
- `make test` — 2,050/2,050 passed through the required Docker path
- `cargo nextest run -p stax-gui` — 178/178 passed
- `cargo nextest run --lib cache::tests::` — 34/34 passed
- final whole-branch review — no remaining findings

## Documentation impact

No README, command documentation, or `skills.md` update is needed. This changes internal cache performance and makes the existing cached-diff behavior responsive; it does not add or change a command, flag, or workflow.

### DIFF
--- a/crates/stax-gui/src/hydration.rs
+++ b/crates/stax-gui/src/hydration.rs
@@ -257,17 +257,20 @@ mod tests {
 
         let session = RepositorySession::open(temp.path()).unwrap();
         let actual = session.diff("feature", "main").unwrap();
-        let cache_path = temp.path().join(".git/stax/tui-diff-cache.json");
+        let cache_dir = temp.path().join(".git/stax/diff-cache/v1");
+        let mut cache_entries = fs::read_dir(&cache_dir)
+            .unwrap()
+            .map(|entry| entry.unwrap().path())
+            .filter(|path| {
+                path.extension().and_then(|extension| extension.to_str()) == Some("json")
+            })
+            .collect::<Vec<_>>();
+        assert_eq!(cache_entries.len(), 1);
+        let cache_path = cache_entries.pop().unwrap();
         let mut stored: serde_json::Value =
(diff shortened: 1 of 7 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #607**
    * **PR #611**
      * **PR #612**
        * **PR #615**
          * **PR #616**
            * **PR #617** 👈
              * **PR #618**

PR #629, opened 2026-07-15
Fix worktree promotion for nested worktrees

cesarferreira:
## Motivation

Fix cases where `stax wt promote` could misidentify the current worktree when a linked worktree lives underneath another worktree path, causing the wrong entry to be treated as current.

- Detect the deepest matching worktree path before marking the current worktree, so nested worktrees are handled correctly.
- Add a regression test covering promotion from a nested linked worktree under `.worktrees/`.

### DIFF
--- a/src/git/repo.rs
+++ b/src/git/repo.rs
@@ -580,6 +580,16 @@ impl GitRepo {
             &mut current_prunable_reason,
         );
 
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #629** 👈

PR #642, opened 2026-07-17
refactor(forge): introduce Forge trait seam and neutral model module

cesarferreira:
## Motivation

Bottom of a 5-PR architecture-deepening stack (#642 → #646) from a codebase review. The forge abstraction was an `enum` + `dispatch!` macro rather than a real seam: there was no way to substitute a fake, so every forge test had to stand up a wiremock HTTP server. Worse, the *forge-neutral* domain types (`PrInfo`, `CiStatus`, `PrMergeStatus`, …) were defined inside the `github` module and imported by GitHub-named paths across ~15 files, and `ci.rs` reached past the seam into `client.octocrab`.

## Description of changes / notes for reviewers

- Add `trait Forge` (`src/forge/traits.rs`) with the 37 forge operations, implemented by `GitHubClient`, `GitLabClient`, `GiteaClient`, and `ForgeClient`.
- Move the forge-neutral model out of `src/github/` into `src/forge/model.rs`, with backward-compatible `pub use` re-exports that keep existing public and import paths working.
- Wire `application::pull_request::resolve_with_lookup` through `Forge`, removing the duplicate one-method `PullRequestLookup` seam.
- Add a configurable `FakeForge` test double and exercise the pull-request fallback path without a mock HTTP server.
- `ForgeClient` stays an enum but now delegates to the trait — lowest-churn shape.

**Not in scope (deliberate):** GitLab/Gitea stubs and the existing `fetch_checks` behavior are preserved verbatim (the CI layer inversion is fixed later in the stack, #644).

**Verification:** pure refactor, no user-visible behavior change. Full suite green under Docker (`make test`); `make lint` clean; forge and pull-request targeted tests pass.

**Risk:** low. Trait/inherent method resolution relies on inherent-method priority (no recursion); the fake-backed application test and existing wiremock HTTP tests exercise both injected and concrete paths.

PR #643, opened 2026-07-17
refactor(merge): consolidate duplicated helpers across the merge family

cesarferreira:
## Motivation

Part of the architecture-deepening stack (#642 → #646). `stax merge` fans out to five sibling command modules that each re-implemented the same wait/scope/display/notification logic: `wait_for_pr_ready` existed 3×, `display_width` 5×, `WaitResult` 4×, `send_notification` 3×, and the scope rule was tested three separate times because it existed three times — while `merge_remote` and `merge_queue` had **zero** tests.

## Description of changes / notes for reviewers

- Add `src/commands/merge_shared.rs` consolidating the duplicated helpers into one copy: `WaitResult`, `wait_for_pr_ready` (blocked-message style parameterized so each command keeps its exact text), `record_ci_history_for_branch`, `strip_ansi`/`display_width`, the `print_header` family, a shared `Scope` + `calculate_scope`, and the PR-retarget cluster (moved off the `merge` command so siblings depend on a neutral module instead of reaching sideways into it).
- Move desktop notifications into `src/notifications.rs` next to sound playback, with a pure `notification_script` seam + escaping test.
- Add scope unit tests for `merge_remote` and `merge_queue` (previously untested paths).

**Not in scope (deliberate):** a shared merge-loop *engine* was declined — `merge_remote` (API-only) and `merge_when_ready` (local rebase/push) have genuinely different loop bodies and user-facing output, so parameterizing them would risk behavior drift. `merge_rebase.rs`, CLI dispatch, flags, and every user-facing string are unchanged.

**Verification:** pure refactor, no behavior change. Full suite green under Docker (2084 pass); the four `print_header*` copies were confirmed byte-identical before consolidating; `blocked_reason` (detailed) vs `status_text` behavior preserved per-caller via a style param.

**Risk:** low–moderate (touches all five merge modules). Highest-attention area is the unified `wait_for_pr_ready` message text — covered by relocated `test_blocked_reason_*` tests.

PR #644, opened 2026-07-17
refactor(ci): fix the CI layer inversion

cesarferreira:
## Motivation

Part of the architecture-deepening stack (#642 → #646). The CI dependency gradient was inverted for the GitHub arm only: `Forge::fetch_checks`' GitHub branch called *up* into `crate::commands::ci::fetch_github_checks`, so the lowest abstraction (raw GitHub check-run HTTP, hitting `client.octocrab`) physically lived in the top command file and the dependency ran `application → forge → commands`. GitLab/Gitea already kept fetch inside their adapters.

- Move GitHub check-run fetching down into the adapter: new `src/github/checks.rs` (`impl GitHubClient`) so the `Forge` arm calls its own inherent method — symmetric with the GitLab/Gitea arms. The `application → forge → commands` upward call is gone.
- Move the provider-neutral `CheckRunInfo` transforms (dedup, commit-status normalization) into `src/ci/normalize.rs` so fetch and render share one copy.
- Relocate the affected unit tests alongside the moved code.

**Not in scope (deliberate):** the further split of `commands/ci.rs` (network / render / watch) is deferred — its ~1,000-line test module is too entangled to split safely for zero behavior benefit. Depends on the `Forge` trait from #642.

**Verification:** pure refactor, no behavior change. Full suite green under Docker (2083 pass). Specifically confirmed: no infinite recursion (`self.fetch_checks` resolves to the inherent method — validated by green wiremock tests, not a hang), and the `application::ci` redirect/auth-stripping + caching tests pass with the check-runs-then-statuses request order preserved.

**Risk:** low. Fetch bodies moved verbatim (only `client.*` → `self.*`); octocrab client construction (where auth-stripping lives) untouched.

### DIFF
--- a/src/ci/mod.rs
+++ b/src/ci/mod.rs
@@ -1,4 +1,5 @@
 pub mod history;
+pub mod normalize;
 
 use serde::{Deserialize, Serialize};
 
--- a/src/ci/normalize.rs
+++ b/src/ci/normalize.rs
@@ -1,0 +1,309 @@
+use super::CheckRunInfo;
+use super::history;
+use crate::git::GitRepo;
+use chrono::{DateTime, Utc};
+use serde::Deserialize;
+use std::collections::HashMap;
+
+/// Response from the commit statuses API
+#[derive(Debug, Deserialize)]
+pub(crate) struct CommitStatus {
+    pub context: String,
+    pub state: String,
+    pub target_url: Option<String>,
+    pub created_at: Option<String>,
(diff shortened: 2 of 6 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #642**
    * **PR #643**
      * **PR #644** 👈
        * **PR #645**
          * **PR #646**

PR #645, opened 2026-07-17
refactor(submit): unify branch-selection into one canonical function

cesarferreira:
## Motivation

Part of the architecture-deepening stack (#642 → #646). Branch selection for `submit` was implemented twice — `commands/submit.rs::resolve_branches_for_scope` and `application/submit.rs::branches_for_submit_scope` — so the two could silently drift, and a scope bug could exist in one without the other's tests noticing.

- Make `application/submit.rs::branches_for_submit_scope` the single canonical selector; `resolve_branches_for_scope` now delegates to it (identical signature, both callers untouched).
- Add 6 unit tests pinning selection for Branch / Downstack / Upstack / Stack scopes plus edge cases, including a regression test asserting the two entry points agree.

The two implementations were proven to produce **byte-identical output on every reachable input** before unifying (the application version's extra dedup/`retain` steps are provable no-ops given `Stack::ancestors/descendants/current_stack` semantics), so this changes nothing at runtime.

**Not in scope (deliberate):** the full `submit`/`sync` `plan()`/`apply()` split remains a larger follow-up — it's not a mechanical refactor and is tied to the strategic decision of whether to commit to (or delete) the half-migrated `application/` layer (worth an ADR).

**Verification:** pure refactor, no behavior change. Full suite green under Docker (2089 pass); the 6 new tests pass including the `resolve_and_application_selection_match` regression pin.

**Risk:** very low. Single delegation + additive tests; no signature or control-flow change.

### DIFF
--- a/src/application/submit.rs
+++ b/src/application/submit.rs
@@ -1029,7 +1029,7 @@ fn prepared_request_verify_hooks(prepared: &PreparedSubmit) -> bool {
     prepared.verify_hooks
 }
 
-fn branches_for_submit_scope(
+pub(crate) fn branches_for_submit_scope(
     stack: &Stack,
     current_branch: &str,
     scope: SubmitScope,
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #642**
    * **PR #643**
      * **PR #644**
        * **PR #645** 👈
          * **PR #646**

PR #648, opened 2026-07-18
fix: preserve auth for gh-stack probes

### LINKED ISSUE #647: st doctor misreports installed gh-stack as missing with env-only GitHub auth
## Summary

`st doctor` reports that `github/gh-stack` is missing even though the extension is installed and runnable when GitHub CLI is authenticated only through `GH_TOKEN` / `GITHUB_TOKEN` and has no stored OAuth login.

This also makes default native-stack auto-registration skip an otherwise valid installed extension.

## Reproduction

Environment:

- macOS
- stax 0.96.7 (also reproduced with 0.94.0)
- gh 2.96.0
- github/gh-stack 0.0.8
- `GH_TOKEN` and/or `GITHUB_TOKEN` set
- no keychain-backed `gh auth login`

```console
$ gh extension list
gh stack  github/gh-stack  v0.0.8

$ gh stack --version
gh stack version 0.0.8

$ st doctor
...
⚠ GitHub native stacks: gh-stack extension missing (run `gh extension install github/gh-stack`)
...

$ gh extension install github/gh-stack
! Extension github/gh-stack is already installed
```

The underlying difference is visible by reproducing the environment used by stax:

```console
$ env -u GH_TOKEN -u GITHUB_TOKEN gh extension list
To get started with GitHub CLI, please run:  gh auth login
...
$ echo $?
4

$ env -u GH_TOKEN -u GITHUB_TOKEN gh stack --version
gh stack version 0.0.8
```

## Cause

`extension_status_with_env` runs `gh extension list` through `gh_command`. `gh_command` removes `GH_TOKEN` and `GITHUB_TOKEN` for every GitHub CLI invocation. The non-zero auth result from `gh extension list` is then classified as `ExtensionStatus::NoExtension`.

Token removal is necessary for the private-preview `gh stack link` / `unstack` API calls, but local extension discovery does not need that auth override removed.

## Expected behavior

`st doctor` should distinguish these states:

- extension installed
- extension missing
- OAuth login missing for native stack API access

An installed extension should not be reported as missing merely because the OAuth prerequisite is absent.

## Suggested minimal fix

Preserve `GH_TOKEN` / `GITHUB_TOKEN` for local capability probes such as:

- `gh --version`
- `gh extension list`
- `gh stack --help`
- installed extension version detection

Strip them only for native-stack remote operations (`gh stack link` and `gh stack unstack`). Alternatively, classify GitHub CLI exit code 4 as an auth/setup failure rather than `NoExtension`.

Regression coverage could use a fake `gh` that lists `github/gh-stack` only while an env token is present, while separately asserting that `link` and `unstack` still receive neither token variable.

geoHeil:
## Summary

- preserve `GH_TOKEN`/`GITHUB_TOKEN` for local `gh-stack` discovery, install, and upgrade probes
- strip PAT overrides only from native stack API operations that require keyring OAuth
- detect Nix-managed `gh-stack` installs by their `gh stack link` capability when `gh extension list` cannot report repository metadata
- add regressions for env-only auth and metadata-free extension installs

Closes #647.

## Testing

- `cargo nextest run gh_stack_tests::` (37 passed)
- `make lint-fast`
- full Docker suite delegated to CI after local Docker DNS failures while building the test image

## Documentation

Not updated: this fixes extension/auth detection and restores the already documented native-stack behavior; it does not change the CLI or configuration contract.

### DIFF
--- a/src/github/gh_stack.rs
+++ b/src/github/gh_stack.rs
@@ -90,12 +90,11 @@ fn extension_status_with_env(env: &[(&str, &str)]) -> ExtensionStatus {
     match extensions {
         Ok(output) if output.status.success() => {
             let stdout = String::from_utf8_lossy(&output.stdout);
-            if stdout.lines().any(|line| line.contains("github/gh-stack")) {
-                if link_command_supported(env) {
-                    ExtensionStatus::Installed
-                } else {
-                    ExtensionStatus::Outdated
-                }
+            let extension_listed = stdout.lines().any(|line| line.contains("github/gh-stack"));
+            if link_command_supported(env) {
+                ExtensionStatus::Installed
+            } else if extension_listed {
+                ExtensionStatus::Outdated
             } else {
                 ExtensionStatus::NoExtension
             }
@@ -233,7 +232,7 @@ pub fn link_stack_with_env(
     remote: &str,
     env: &[(&str, &str)],
 ) -> LinkOutcome {
-    let mut command = gh_command(env);
+    let mut command = gh_stack_command(env);
     command.args(["stack", "link"]);
     for number in pr_numbers {
         command.arg(number.to_string());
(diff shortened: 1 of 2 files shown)


cesarferreira (approved):
The command split is correct: local discovery/install operations retain environment authentication, while remote stack operations continue removing PAT overrides. CI is fully green and `git diff --check` passes.

Should-fix:
- `src/github/gh_stack.rs:278,299,319` — The authentication boundary is only partially covered. Tests verify token preservation through `doctor` and removal for `link`, but not removal for `unstack` or preservation during install/upgrade. Given the repository’s testing policy, add fake-`gh` assertions for those paths.

Question:
- `tests/gh_stack_tests.rs:189` — Issue #647 also asks `doctor` to distinguish “extension installed” from “OAuth login missing.” This test only verifies the former, so the exact reproduction now reports the extension accurately but still gives no OAuth warning. Is that intentionally deferred to a follow-up?

No blockers found.

PR #656, opened 2026-07-20
feat(tui): add Emacs-style key bindings across all TUIs

kcolford:
Introduce a shared `tui::keys` module that normalizes Ctrl-key shortcuts into the plain keys the existing match arms understand, and provides a common line-editor for text-entry fields.

- Motion (Ctrl+N/P/F/B) maps to Down/Up/Right/Left in every context, including text-entry modes where a bare n/p is literal text.
- Line editing (Ctrl+A/E/D/K, Ctrl+G to cancel) applies only while typing, so single-letter shortcuts in list views stay intact.
- Ctrl+C quit and Ctrl+P survive a stray Shift modifier reported by some terminals.

Wired through the main dashboard, ready, split, split_hunk, and worktree TUIs, replacing each hand-rolled cursor/edit block with the shared helper. Help modals updated to document the new bindings.


cesarferreira:
appreciate the PR but it feels very opinionated towards emacs users, why do you think this is a good idea?

PR #659, opened 2026-07-20
feat(resolve): retry AI conflict resolution with feedback on invalid output

kcolford:
## What

`stax resolve` invokes an external AI CLI agent to resolve rebase conflicts and expects JSON back. Previously **any** malformed JSON or schema-rule violation aborted the whole command. This wraps the invoke→parse→validate step in a bounded **repair loop** (`MAX_PARSE_ATTEMPTS = 3`): on a schema/validation failure it feeds the rejection reason and the agent's prior output back through a new repair prompt so the agent can self-correct.

## Scope for reviewers

- **`src/commands/resolve.rs`** — the whole change lives in the new `obtain_resolutions` + `build_repair_prompt`. Worth a close read:
  - Only **schema/validation** failures retry; a failed agent *invocation* still aborts immediately (via `?`).
  - The repair loop is orthogonal to the existing `--max-rounds` git-conflict loop — attempts live entirely inside one round.
  - Happy path borrows the base prompt via `Cow<str>` to avoid cloning the full conflicted-file payload.
- **`tests/resolve_tests.rs`** — `test_resolve_retries_invalid_response_then_succeeds` proves the loop: fake agent emits bad JSON first, valid JSON on retry.

## Not in scope (noted)

The pre-existing `strip_markdown_fences` (resolve) vs `generate::extract_ai_json_blob` duplication is the natural next generalization target (a shared `invoke_ai_agent_json::<T>()`), but it's outside this diff.

## Testing

- `cargo check` clean, no new clippy warnings in `resolve.rs`
- All 7 `resolve_tests::` pass

### DIFF
--- a/src/commands/resolve.rs
+++ b/src/commands/resolve.rs
@@ -4,6 +4,7 @@ use crate::git::{GitRepo, RebaseResult};
 use anyhow::{Context, Result, bail};
 use colored::Colorize;
 use serde::Deserialize;
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::fs;
 use std::io::IsTerminal;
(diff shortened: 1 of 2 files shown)


cesarferreira:
Thank you for the PR, do you mind solving the CI check failure so we can merge it?

PR #664, opened 2026-07-21
fix(worktree): handle external relative roots

### LINKED ISSUE #654: feat: allow global configuration of worktree structure
### Context
Currently, I keep every repo in a consistent on-disk layout: a plain container folder (not a git repo) that holds the trunk clone plus every worktree lane as siblings:

```
my-repo/             <- container (plain folder; shared .env* live here)
  development/       <- the trunk clone (Stax home)
  feature-x/         <- st wt create -> lane here
  feature-y/         <- lane
```

Ideally, I'd like to configure this once, globally (in `~/.config/stax/config.toml`) and have it apply to every repo I clone this way.

### Problem
A relative `root_dir` already does exactly the right resolution, globally:

```toml
[worktree]
root_dir = ".."   # lane resolves to <clone>/../<lane> = the container
```

This is repo-agnostic (it names no specific path), so one global line places lanes correctly for every repo, regardless of the clone's name.

The problem: with a **relative** `root_dir`, `stax worktree create` appends the raw relative path as an ignore pattern (`../`) to the repo's `.gitignore` — even though the lane is created **outside** the clone's working tree and is never at risk of being tracked. On a repo with a committed `.gitignore`, that dirties a tracked file:

```bash
# repo with a tracked .gitignore, clean working tree, relative root_dir = ".."
stax worktree create mylane --no-verify

$ git status --porcelain
 M .gitignore
$ git diff -- .gitignore
+../
```

The added `../` is also a no-op — `git check-ignore -v <anyfile>` → exit 1, matches nothing. So on shared repos I get an unwanted `.gitignore` change in `git status`/PR diffs for zero benefit. (Reproduced on stax **0.96.7**.)

### Current workaround
A per-repo local `stax.toml` with an absolute `root_dir`, hidden via `.git/info/exclude`:

```toml
[worktree]
root_dir = "/abs/path/to/my-repo"
```

An absolute `root_dir` makes stax skip the `.gitignore` write, so `git status` stays clean. But it's absolute, so I have to create and maintain one of these files per repo, which is exactly the manual, per-repo overhead I would like to avoid with a possible global rule.

### Suggestion
Support a global worktree-structure config (container-with-sibling-lanes) that doesn't require a per-repo file and doesn't dirty tracked files.

cesarferreira:
## Summary

- resolve configured worktree roots before deciding whether `.gitignore` needs an entry
- keep external relative roots such as `..` from dirtying tracked files
- document the global sibling-lane layout and cover internal, external, and root-equality cases

## Testing

- `make lint`
- `make test` (2,103 passed)

Fixes #654

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -79,6 +79,7 @@ Config is loaded as follows:
 
 [worktree]
 # root_dir = "" # default: ~/.stax/worktrees/<repo>
+# root_dir = ".." # optional: keep lanes beside a nested trunk clone
 # reuse_slots = true
 #   Warm-slot recycling. Removing a clean, merged-equivalent worktree parks it
 #   (reset --hard trunk + `git clean -fd`, keeping gitignored deps like
--- a/docs/worktrees/index.md
+++ b/docs/worktrees/index.md
@@ -216,6 +216,7 @@ Override in `~/.config/stax/config.toml`, or set shared project overrides in rep
 [worktree]
 # root_dir = ""             # default external root
 # root_dir = ".worktrees"   # keep worktrees inside the repo
+# root_dir = ".."           # keep lanes beside the main checkout
 
 # Warm-slot recycling: a removed worktree is parked and reused by the next lane
 # instead of being deleted, keeping built gitignored deps on disk.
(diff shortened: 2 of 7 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #664** 👈

PR #682, opened 2026-07-23
fix(restack): repair stale parent metadata after external rebase

### LINKED ISSUE #679: bug: Restack can replay stale commit ranges after an external rebase
## Description

`st restack` can replay an outdated commit range when a branch has already been rebased onto its parent but Stax metadata still records the previous parent revision.

This can reintroduce commits that were already squash-merged, and repeated `st restack` invocations can repeat the same rebase.

Observed with Stax 0.98.0.

## Reproduction

1. Create a tracked feature branch containing several commits.
2. Advance its parent branch.
3. Rebase the feature branch using Git directly:

   ```sh
   git rebase main
   ```

   This updates the Git graph but leaves Stax's stored parent revision stale.

4. Run:

   ```sh
   st restack
   ```

## Expected behavior

Stax should recognize that the branch is already based on the current parent tip, repair its stored parent revision, and avoid replaying commits.

## Actual behavior

Restack may use the stale stored revision as the replay boundary. Commits from the outdated range can be replayed, including commits already represented on the parent through squash merges.

## Root cause

The restack preflight only treats the stored boundary as suspicious when the replay range exceeds its size/ratio heuristic.

It misses the stronger structural signal:

- the current parent tip equals the branch merge-base; and
- the stored parent revision differs from that tip.

In this state, the branch is already based on the current parent and the stored revision is stale regardless of replay-range size.

## Suggested fix

When the parent tip equals the merge-base but differs from the stored parent revision, treat the metadata as stale and use the merge-base as the corrected upstream boundary.

A regression test can:

1. Create a feature branch with two commits.
2. Advance `main`.
3. Externally rebase the feature onto `main`.
4. Preserve the old Stax parent revision.
5. Run `st restack`.
6. Assert that the branch SHA and its two feature commits remain unchanged and the metadata no longer reports a restack requirement.

cesarferreira:
## Motivation

Fixes #679

`st restack` could replay an outdated commit range when a branch had already been rebased onto its parent externally (e.g. `git rebase main`), but stax metadata still recorded the previous parent revision. This could reintroduce commits already squash-merged into the parent, and repeated `st restack` invocations could repeat the same no-op-that-wasn't rebase.

- The existing restack preflight only treated the stored parent revision as suspicious when the replay range exceeded a size/ratio heuristic (`is_suspicious()` in `src/engine/restack_preflight.rs`). It missed the stronger structural signal: when the parent's current tip equals the branch's merge-base but differs from the stored parent revision, the branch is already based on the current parent regardless of replay-range size.
- Added `RestackPreflight::already_rebased_externally()` to detect this case, and widened `corrected_upstream()` to use the merge-base as the boundary when either the size heuristic or the new structural signal fires. `choose_rebase_upstream()` now emits a distinct diagnostic reason for the structural case.
- Added a regression test (`tests/restack_provenance_tests.rs`) that reproduces the exact repro from the issue: create a feature branch, advance main, externally rebase the feature onto main, preserve the stale stax parent revision, then run `st restack` — asserts the branch SHA and its commits are unchanged and the metadata self-repairs so no further restack is reported as needed.

## Testing

- `cargo check && cargo clippy -- -D warnings`
- `cargo nextest run restack_preflight::` (9 passed)
- `cargo nextest run restack_provenance_tests::` (22 passed, including the new regression test)

### DIFF
--- a/src/engine/restack_preflight.rs
+++ b/src/engine/restack_preflight.rs
@@ -10,6 +10,13 @@
 //! boundary is clearly worse than the current merge-base. The correction is
 //! intentionally conservative and only affects the current rebase invocation;
 //! normal restack success still refreshes metadata to the parent tip.
+//!
+//! Besides the count-based heuristic, the module also detects a *structural*
+//! signal: when the branch has already been rebased externally so that the
+//! parent tip equals the merge-base but differs from the stored boundary, the
+//! stored boundary is stale and replaying from it would re-apply commits that
+//! are already merged. In that case restack rebases from the merge-base (a
+//! no-op) rather than the stale stored boundary.
 
 use crate::config::Config;
 use crate::git::GitRepo;
@@ -39,6 +46,10 @@ pub struct RestackPreflight {
     /// `stored_revision`.
     #[allow(dead_code)]
     pub merge_base: Option<String>,
+    /// Current tip of `parent` when available. Used to detect an external
+    /// rebase that left the stored boundary stale.
+    #[allow(dead_code)]
+    pub parent_tip: Option<String>,
     pub stored_to_branch: Option<usize>,
     pub merge_base_to_branch: Option<usize>,
 }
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #682** 👈

PR #695, opened 2026-07-28
sync: warn when a merged/closed-PR branch has unpushed commits

cesarferreira:
## Summary

`st sync` (alias `st rs`) silently skips branches it decides not to delete, even when the reason is that the branch's PR is merged/closed but the branch still carries local commits that exist nowhere else. This surfaces that case instead of staying silent.

When "Detect merged branches" spares a branch for this reason, sync now prints a note like:

```
⚠ cesar/foo: PR #693 merged, but branch has 5 additional commits not on trunk or any remote — not deleting
```

## Changes

- `src/commands/sync.rs`: added `PartialMergeReason` (PrMerged / PrClosed / HistoryMerged), `PartiallyMergedNote`, and `find_partially_merged_notes()` — runs alongside the existing `find_merged_branches()` using only local metadata (no GitHub API calls). Only fires when the branch actually has commits missing from trunk and every remote (`count_extra_commits > 0`), so fully-pushed branches stay silent as before.
- Purely additive: deletion logic (`find_merged_branches`, the delete loop) is untouched — this only changes what gets printed.
- `tests/integration_tests.rs`: added `test_sync_notes_closed_pr_with_extra_local_commits` and `test_sync_no_note_when_branch_fully_pushed` to cover the new note and its conservative gate.

## Verification

- `cargo check`, `cargo clippy -- -D warnings`, `cargo fmt --check` — all clean
- Targeted: `cargo nextest run test_sync_notes_closed_pr_with_extra_local_commits test_sync_no_note_when_branch_fully_pushed test_sync_does_not_treat_closed_unmerged_pr_as_merged test_sync_deletes_merged_branches test_sync_preserves_unmerged_branches` — 5/5 passed
- Full suite: `make test` — 2149/2149 passed

### DIFF
--- a/src/commands/sync.rs
+++ b/src/commands/sync.rs
@@ -642,6 +642,16 @@ pub fn run(
                 .as_ref()
                 .expect("remote branch list when deleting merged branches"),
         )?;
+        let partially_merged_notes = find_partially_merged_notes(
+            &repo,
+            &workdir,
+            &stack,
+            &remote_name,
+            remote_branches_for_merged
+                .as_ref()
+                .expect("remote branch list when deleting merged branches"),
+            &merged,
+        )?;
         step_timings.push((
             "detect merged branches".to_string(),
             detect_merged_started_at.elapsed(),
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #695** 👈

PR #709, opened 2026-07-29
feat(submit): open a PR via a fork when you lack write access

kcolford:
## What

`stax branch submit` previously aborted with a raw git error when the push to the upstream remote was rejected for lack of write access. This adds the common contributor path: fall back to submitting the branch **from a fork**.

When a push is rejected and the user opted in (`--fork` or `remote.auto_fork = true`):
- classify the rejection as a permission denial (conservatively — never a non-fast-forward reject);
- reuse a pushable fork under the user's login, or create one via the GitHub API;
- ensure a `fork` git remote (URL scheme matched to origin) and push the branch there, retrying with backoff for a freshly created fork;
- open the PR with head `<fork_owner>:<branch>`, base unchanged, `maintainer_can_modify = true`.

## Scope

**Single branch only.** A multi-branch stack against an upstream you can't push to errors clearly — a stacked child PR's base branch can't live in the upstream repo. Not-opted-in and missing-base cases also error actionably. GitHub only; GitLab/Gitea cleanly reject fork mode.

## Where to look

- `src/application/submit.rs` — `push_error_indicates_no_write_access` classifier (the false-positive guard is the critical bit; unit-tested).
- `src/commands/submit.rs` — fallback wiring at the push/create_pr sites in `run`.
- `src/github/client.rs` — `find_pushable_fork` / `create_fork`.
- `src/git/repo.rs` — `ensure_remote`, `push_branch_to_fork`.
- Config `remote.auto_fork` / `remote.fork_remote`; CLI `--fork`.

## Follow-ups (not in this PR)

`--no-maintainer-edit`; fork-aware `find_open_pr_by_head` for *untracked* pre-existing fork PRs; GitLab/Gitea support.

## Tests

New unit tests (classifier, fork lookup/create) + 3 integration tests (fork push+PR, not-opted-in error, multi-branch error). `make test` green; `cargo check`/`fmt` clean; no new clippy findings vs baseline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,7 @@ st config --set-ai
 | `st ss` | Submit the full stack, open/update linked PRs; temporary-publishes branches that need restack |
 | `st submit --plan [--json]` | Preview fetch, push, PR, retarget, metadata, and stack-link actions without changing local or remote state |
 | `st branch submit` | Submit only the current branch; can publish a temporary rebased head when needed |
+| `st branch submit --fork` | On a denied-for-lack-of-write-access push, retry from a GitHub fork of the upstream repo (single branch; set `[remote] auto_fork = true` for always-on) |
 | `st upstack submit` | Submit current branch and descendants, chaining temporary publish heads when needed |
 | `st reviews --stack [--json]` | Stack-wide review/comment inbox, including inline file/line locations on GitHub (`st comments` remains the current-PR view) |
 | `st next` | Move to the next unmerged branch upstack; fork choices are deterministic |
--- a/docs/commands/reference.md
+++ b/docs/commands/reference.md
@@ -282,9 +282,10 @@ If the stash cannot apply cleanly while committing below, Stax restores the orig
 - `--template <name>` / `--no-template` / `--edit`
 - `--rerequest-review` / `--update-title`
 - `--native-stack` force-attempt native GitHub Stack registration for this submit; `--no-native-stack` skips it
+- `--fork` if the upstream push is denied for lack of write access, fall back to submitting the branch from a fork (single branch, GitHub only)
 - `--yes` / `--no-prompt`
 
-Config: `[submit] stack_links = "comment" | "body" | "both" | "off"` and `native_stack = "auto" | "off" | "link"` in `~/.config/stax/config.toml`.
+Config: `[submit] stack_links = "comment" | "body" | "both" | "off"` and `native_stack = "auto" | "off" | "link"`; `[remote] auto_fork` and `[remote] fork_remote` control the fork fallback — see [Configuration → Fork fallback for submit](../configuration/index.md#fork-fallback-for-submit).
 
 ### `st completions`
 
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -32,6 +32,8 @@ Config is loaded as follows:
 # base_url = "https://github.com"
 # api_base_url = "https://github.company.com/api/v3"
 # forge = "github" # "github" | "gitlab" | "gitea" — override auto-detection
+# auto_fork = false # always fall back to submitting from a fork when the upstream push is denied
+# fork_remote = "fork" # name of a pre-configured git remote for the fork (skip auto-detect)
 
 [submit]
 # stack_links = "comment" # "comment" | "body" | "both" | "off"
@@ -239,6 +241,27 @@ forge = "gitlab"
 Accepted values: `"github"`, `"gitlab"`, `"gitea"`, `"forgejo"` (Forgejo is treated as Gitea).
 
 Auto-detection fallback: hostnames containing `gitlab` → GitLab, `gitea`/`forgejo` → Gitea, otherwise → GitHub.
+
+## Fork fallback for submit
+
+When `stax branch submit` pushes to an upstream you lack write access to, GitHub rejects the push. Opt in to a fork-based fallback so stax re-runs the push against your fork instead:
+
+```toml
+[remote]
+auto_fork = true          # always fall back to a fork on permission-denied push
+fork_remote = "fork"      # optional: reuse an existing git remote you added yourself
+```
+
+Or use `stax branch submit --fork` for a one-off. When enabled:
+
+- stax reuses a pushable fork under your GitHub login, or creates one via the GitHub API when none exists.
+- the branch is pushed with `--force-with-lease` (no bare `--force`); a diverged fork branch fails actionably rather than silently overwriting it.
+- the PR is opened with head `<fork_owner>:<branch>` and `maintainer_can_modify = true`.
+- an existing git remote named `fork` is never silently repointed; the run fails if its URL conflicts.
+
+Supported forges: GitHub only. Fork fallback rejects GitLab/Gitea cleanly.
+Scope: single branch only. A multi-branch stack cannot be submitted from a fork because a stacked child PR's base branch cannot live in the upstream repo.
+The base branch must already exist upstream.
 
 ### Automatic CI hydration trust
 
--- a/skills.md
+++ b/skills.md
@@ -238,6 +238,7 @@ stax submit --ai --yes             # Accept generated new-PR details
 stax submit --rerequest-review     # Re-request existing reviewers on update
 stax submit --native-stack         # Force-attempt native GitHub Stack registration for this run
 stax submit --no-native-stack      # Skip native GitHub Stack registration for this run
+stax submit --fork                 # Fall back to fork on permission-denied push (single branch, GitHub only)
 stax completions zsh               # Generate completions: bash|zsh|fish|powershell|elvish
 
 # ~/.config/stax/config.toml; repo-root stax.toml overlays shared values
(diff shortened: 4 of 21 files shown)


geoHeil (changes requested):
Thanks for tackling the contributor fork workflow. The direction is useful, but I do not think this is safe to merge yet: the happy path is covered, while several realistic paths can overwrite or publish the wrong work.

Blocking findings:

- `src/git/repo.rs:1985`: `git push --force` is not safe for a reused fork. Another checkout or contributor tool may have advanced that branch, and this silently overwrites it. Please reuse the existing `push_branches` lease semantics after resolving the fork branch OID, or otherwise use an explicit `--force-with-lease`.
- `src/commands/submit.rs:1656`: the fallback pushes the local branch name instead of `PushSpec::source_ref`. A scoped submit using a temporary publish ref will therefore publish the wrong commit. This duplicate push path also drops `--no-verify`. Reusing `push_branches` would fix both and delete the new weaker push primitive.
- `src/application/submit.rs:706`: the classifier is not as conservative as its comment claims. A non-fast-forward rejection for a branch named `fix-403` contains `403` in the formatted error and is misclassified as a permission denial. Match the concrete transport phrase such as `returned error: 403`, or classify structured stderr before branch names are added. Please add this regression case.
- `src/commands/submit.rs:2191`: `uses_application_default_submit` does not consider `options.fork`, so `stax submit --no-pr --fork` routes to the application path and silently ignores the new option. The integration test comment works around this by choosing `upstack`; that is a sign the routing bug needs a test and fix.
- This adds a user-visible flag and two config keys but updates none of `README.md`, `docs/`, or `skills.md`, contrary to the repository documentation policy.

Would you be open to making the fork seam forge-neutral now? `ForkTarget` belongs in `forge::model`, and `find_pushable_fork` / `create_fork` fit the existing `Forge` trait and dispatch macro. That removes the GitHub type dependency from `forge`, avoids another provider-specific submit branch, and leaves a direct implementation path for GitLab project forks rather than baking GitHub-only ownership/head formatting into the command.

Cleanup and test suggestions:

- Reuse the runtime and `ForgeClient` already created during planning instead of constructing a second pair in the error arm.
- Do not silently repoint an existing remote named `fork`; validate it or fail with an actionable conflict.
- The happy-path fixture authenticates as `test-owner`, which is also the upstream owner. That cannot model a contributor who lacks upstream write access and masks owner/head bugs. Use a distinct `contributor` login and assert `contributor:<branch>`.
- `write_test_config` and the PR create/refresh WireMock payload duplicate helpers already present in scoped-submit tests. Move the small reusable GitHub fixtures into `tests/common` or keep this scenario beside the existing scoped-submit fixtures.
- Add focused coverage for an existing fork branch/lease, a temporary publish ref, the `fix-403` false positive, and the default-submit router. These catch the actual risky edges; more mock JSON does not.

net: -100 lines possible.

kcolford (commented):
Thanks for the detailed pass, @geoHeil — all five blocking items and the design/cleanup asks are addressed in 02188bc. `cargo check`, `cargo clippy --tests -- -D warnings`, and `make test` (2184 tests) are green.

**Blocking**

- `src/git/repo.rs:1985` (unsafe `--force`) — deleted `push_branch_to_fork`. The fork arm reuses `push_branches`, which already emits per-spec `--force-with-lease` and honours `--no-verify`. Lease expectation resolved with `remote::ls_remote_head_oids` and re-resolved on each retry attempt. A diverged fork branch fails actionably (names the fork remote/branch and points at `remote.fork_remote`) rather than silently overwriting.
- `src/commands/submit.rs:1656` (`source_ref` dropped) — same fix. The fork spec is built by cloning `pushed_branches[0]` and overriding only `expected_remote_oid`, so a temporary publish ref (`refs/stax/submit/...`) is used verbatim.
- `src/application/submit.rs:706` (`403` false positive) — classifier now matches the concrete transport phrases (`returned error: 403`, `403 forbidden`, `write access to repository not granted`) and drops the bare `403` substring. New tests cover the `fix-403` non-fast-forward reject (returns `false`) and a real permission denial on a `fix-403`-named branch (returns `true`).
- `src/commands/submit.rs:2191` (`--fork` bypassed by the default router) — `uses_application_default_submit` now takes `fork_enabled`; `run` hoists `Config::load` and passes `options.fork || config.auto_fork()`. Two router unit tests added. The `multi_branch_stack_...` integration test drops the `upstack` workaround and now uses `stack submit --no-pr --fork` directly.
- Docs policy — `README.md`, `docs/commands/reference.md`, `docs/configuration/index.md` (new **Fork fallback for submit** section), and `skills.md` all updated for the flag and both config keys.

**Design (accepted)**

- `ForkTarget` moved to `forge::model`; `find_pushable_fork` / `create_fork` are on the `Forge` trait behind the `dispatch!` macro. GitHub delegates to the inherent methods; GitLab/Gitea `bail!` with the existing "currently only supported on GitHub" message; `FakeForge` returns the usual "unused in fake". `use crate::github::client::ForkTarget` removed from `forge/mod.rs`; the hand-rolled match arms are gone.

**Cleanups**

- Reused runtime/client — the fork arm consumes the planning `rt`/`client` when planning built them; it only constructs a fresh pair when `--no-pr` silently skipped them (fallback for that path).
- Existing `fork` remote — `remote_url("fork")` is checked before `ensure_remote`; a URL mismatch bails with both URLs and the `remote.fork_remote` remedy. New integration test covers this.
- Contributor fixture — the happy-path test now authenticates as `contributor` (distinct from the upstream owner `test-owner`), mocks `GET /repos/contributor/test-repo` → 404, and asserts the PR head is `contributor:<branch>`.
- Configured `fork_remote` owner — the owner is now derived by parsing the configured remote's URL rather than assuming the authenticated login, so a fork under an org gets the correct `owner:branch` head.
- Shared fixtures — extracted `tests/common/github_mock.rs` with `write_stax_config`, `mount_current_user`, `mount_pr_create_and_refresh`. Both `submit_fork_fallback_tests` and `scoped_submit_tests` migrated. Other suites out of scope.

**Follow-ups (as you noted)**

`--no-maintainer-edit`, fork-aware `find_open_pr_by_head` for untracked pre-existing fork PRs, and GitLab/Gitea fork support remain follow-ups.

geoHeil (approved):

cesarferreira (changes requested):
Thanks for this — the feature is well scoped and the second round addressed @geoHeil's blockers properly. A few things I want sorted before merge.

**Blockers**

1. **CI is red and the new tests never ran.** `Integration Tests` fails on `restack_provenance_tests::test_sync_restack_many_trunk_commits_preserves_linear_feature_depth` (`fatal: could not parse HEAD`), which this PR doesn't touch — looks like an unrelated flake. But nextest fails fast, 243 tests were skipped, and the job log has zero hits for `submit_fork_fallback`. All four new integration tests are unverified. Please re-run.

2. **`src/commands/submit.rs:1709` — the lease can't protect anything.** `expected_remote_oid` is resolved from a live `ls_remote_head_oids` and the push happens on the next statement, so the lease always matches and the force-push always succeeds. If a contributor already has a `feature` branch on their fork from earlier work, `--fork` silently overwrites it. That also makes `docs/configuration/index.md:59` ("a diverged fork branch fails actionably rather than silently overwriting it") untrue, and leaves the `push_error_indicates_stale_lease` arm at `:1730` reachable only by a sub-second race. I'd rather see an explicit check: on the first push to a fork branch stax has never written, bail if the fork tip isn't an ancestor of the local tip. (Can't be unconditional — we legitimately force-push after a restack.)

3. **`tests/submit_fork_fallback_tests.rs:223` asserts nothing.** `branch_submit_fork_conflicting_existing_fork_remote_fails` points at `http://127.0.0.1:1` with no MockServer, and the comment says the conflict check fires before any API call — but the fork arm's first action is `get_current_user()` at `submit.rs:1650` and the check is at `:1686`. Against a dead endpoint the run dies well before that. The test then discards the exit status with `let _ =` and asserts the remote URL is unchanged, which is trivially true. So the "don't repoint an existing `fork` remote" fix has no coverage. Mount a server and assert on stderr.

**Should-fix**

- `src/github/pr.rs:520` — `maintainer_can_modify(true)` is unconditional, so *every* GitHub PR stax creates now requests maintainer edits, not just fork ones. That's a global behaviour change inside a fork feature, untested for the normal path. It also collides with the org-fork support added at `:1676`: GitHub won't let you enable maintainer edits on an org-owned fork, so that combination risks a 422. Please gate it on `fork_owner.is_some()`.
- `src/commands/submit.rs:1723` — 3 attempts × 2s ≈ 4s for a freshly created fork. GitHub fork creation is async and can take minutes on large repos, and this is the main path the feature exists for. Poll `GET /repos/{login}/{repo}` with backoff instead.
- `src/commands/submit.rs:1686` — the conflict check runs after `create_fork()` at `:1672`, so a user with a stale `fork` remote gets a real fork created and *then* an abort. Check the remote first.
- `src/commands/submit.rs:1660` — a configured `fork_remote = "myfork"` that doesn't exist yet is silently dropped and a remote named `fork` is created instead. Use the configured name.
- `src/commands/submit.rs:1598` / `:1609` — both bails drop the underlying push error, unlike `:1590`. The user loses the root cause.
- `src/commands/submit.rs:1595` guards `pushed_branches.len() != 1`, but the head rewrite at `:2057` applies to every create, and a plan with `needs_push == false` and no PR is still a create. `plans.len() != 1` is the right quantity.

**Minor**

- `src/application/submit.rs:731` — `contains("(stale info)")` is subsumed by `contains("stale info")`.
- `src/commands/submit.rs:1724` — `eprintln!` here ignores `quiet`, unlike the surrounding code.
- One design question: the classifier can't tell "no write access to the repo" from "not allowed to push this ref", so a protected-branch rejection worded `not authorized` matches at `application/submit.rs:717`. With `auto_fork = true` that turns a policy error into a surprise fork PR. Intended?

cesarferreira:
Love this feature, thanks for putting it together! Since it's been quiet here for a while, I went ahead and applied the requested changes myself on #808 so we don't lose the work. Closing this one in favor of that — appreciate the contribution!

PR #710, opened 2026-07-29
fix(test): make native tests resilient to PATH clobbering

cesarferreira:
## Summary
- Fix native zsh integration tests in `shell_setup.rs` that failed on machines where `/etc/zsh/zshenv` re-imports `PATH` from `/etc/environment` after startup.
- Add shared test helpers that set `__STAX_BIN` (for wrapper behavior tests) or inline `export PATH=…` (for PATH lookup tests) inside the zsh command body, after global shell init.
- Fix Linux integration tests that use `script -qefc` by re-exporting `PATH` inside the recorded command shell, so fake `gh` binaries are not bypassed by the same system PATH reset.
- Standardize zsh shell-setup tests on `zsh -dfc` instead of `-lc` so user shell config and installed `stax` binaries do not interfere.

## Test plan
- [x] `cargo nextest run shell_setup::tests::posix_shell_snippet` (9/9 passed)
- [x] `cargo nextest run doctor_fix_installs_missing_gh_stack_extension doctor_fix_upgrades_outdated_gh_stack_extension` (2/2 passed)
- [x] `make test-native` (2176/2176 passed on Azure dev VM)
- [x] `make lint-fast`

No user-visible behavior changes; docs not updated.

### DIFF
--- a/src/commands/shell_setup.rs
+++ b/src/commands/shell_setup.rs
@@ -948,6 +948,38 @@ mod tests {
     use std::{fs, io::ErrorKind, path::Path, process::Command};
     use tempfile::tempdir;
 
+    #[cfg(unix)]
+    fn zsh_available_with_flags(flags: &str) -> bool {
+        let mut cmd = Command::new("zsh");
+        for flag in flags.split_whitespace() {
+            cmd.arg(flag);
+        }
+        cmd.arg("exit").arg("0");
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #710** 👈

PR #724, opened 2026-07-30
fix(track): set upstream on branches fetched by --all-prs

cesarferreira:
## Motivation

`stax branch track --all-prs` fetches branches with `<b>:<b>` refspecs (`fetch_branches_batch`, `fetch_branch_from_remote`). That form creates the local branch but configures **no** `branch.<b>.remote` / `branch.<b>.merge`, so a plain `git pull` on any branch the command just created fails with:

```
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.
```

Hit in a real monorepo after a `--all-prs` run.

- After fetching, configure the upstream — **only for branches this run actually created** (in `local_after`, absent from `local_before`). A pre-existing local branch may intentionally track a fork or a different remote; rewriting that isn't stax's call, and it isn't undoable.
- Writes `branch.<b>.merge` + `branch.<b>.remote` config directly instead of shelling out to `git branch --set-upstream-to=<remote>/<b>`. Verified on git 2.54: that command **hard-fails** (`fatal: the requested upstream branch 'origin/feat1' does not exist`) when `refs/remotes/<remote>/<b>` is absent, and a `<b>:<b>` refspec fetch only writes the remote-tracking ref *opportunistically*, via the remote's default `remote.<name>.fetch`. Config-only works either way.
- `.merge` is written before `.remote` deliberately: if the second write fails, git defaults an unset `branch.<name>.remote` to `origin`, so the partial state still usually works. The reverse order leaves `git pull` erroring.
- Failures **warn once, aggregated, and never abort** the run. No per-branch line on the happy path; the summary gains a dimmed `Set upstream to 'origin' on N newly fetched branch(es).` only when at least one was set.
- `branches_with_configured_upstream()` treats `git config --get-regexp` exit code 1 as "no matches" (the normal case in a fresh repo), not an error.

New pure logic lives in `src/application/track_plan.rs` alongside the existing `--all-prs` decision logic; the two git-touching methods are on `GitRepo`.

## Tests

- 5 new pure-logic tests in `tests/track_plan_tests.rs` (already-registered file) — covers the conservative-selection guarantee, failed fetches, and config parsing for dotted/slashed branch names like `release.1.x` and `feat/x`.
- 3 new unit tests in `src/git/repo.rs` — the main one builds the branch via a real `git fetch --no-tags origin copied:copied` from a bare remote (mirroring what track does) and asserts `git rev-parse --abbrev-ref copied@{upstream}` resolves to `origin/copied`, i.e. the exact thing the user's `git pull` needs.
- Full suite: 2263 passed, 0 failed. `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/docs/commands/reference.md
+++ b/docs/commands/reference.md
@@ -84,7 +84,7 @@ See also: [Merge and cascade](../workflows/merge-and-cascade.md)
 | `st rename` | | Rename current branch |
 | `st move [target]` | `mv` | Move the current branch and descendants onto a new parent (`st upstack onto` parity alias; picker when omitted) |
 | `st branch track` | | Track an existing branch |
-| `st branch track --all-prs` | | Track all open PRs (GitHub, GitLab, Gitea) |
+| `st branch track --all-prs` | | Track all open PRs (GitHub, GitLab, Gitea); sets upstream on branches it fetches |
 | `st branch untrack` | `ut` | Remove stax metadata |
 | `st branch reparent` | | Change parent |
 | `st branch submit` | `bs` | Submit current branch only; can temporarily restack the publish head when the excluded parent is remote-synced |
--- a/src/application/mod.rs
+++ b/src/application/mod.rs
@@ -37,6 +37,7 @@ pub(crate) use submit::{
     SubmitPromptRequest, SubmitScope,
 };
 pub use track_plan::{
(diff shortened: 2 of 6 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #724** 👈

PR #727, opened 2026-07-30
Let setup and skills update target selected agent harnesses only.

cesarferreira:
## Summary

Addresses #726: `stax setup` no longer installs the stax skill for every agent harness by default.

- Interactive setup: after the skills yes/no prompt, a multi-select lists Codex, OpenCode, Claude, Cursor, and pi — pre-checked when the harness directory exists on disk.
- Non-interactive: `st setup --skills all|detected|auto|none|claude,codex,...` (composable with `--install-skills`).
- `st setup --yes` now installs skills only for **detected** harnesses (use `--skills all` for the previous behaviour).
- Persists `[skills] harnesses` in `~/.config/stax/config.toml` so `st skills update`, `st skills list`, and `st doctor` respect the selection; `st skills update --all` overrides.

## Test plan

- [x] `make lint` / `make test`
- [x] Unit tests for harness selection parsing
- [x] Integration tests for setup `--skills`, `--yes` detection, config-scoped update, and `--all` override

PR #728, opened 2026-07-31
fix(restack): drop become-empty commits instead of erroring

cesarferreira:
## Problem

`st restack` failed with:

```
Error: Could not finish rebasing 'cesar/XXX-2943-...'
Inspect the rebase state, then run `st continue` or `st abort`
git rebase stopped for 'cesar/XXX-2943-...' without unresolved file conflicts
```

The rebase hit a commit that became **empty** — its changes were already present in the new base. This happens routinely after a parent PR squash-merges into trunk, or when the same change lands upstream independently. Git stops such a rebase and leaves it in progress **without any file conflict**.

stax misclassified that stopped-but-conflict-free state as a failure and bailed. Worse, the suggested recovery was also broken: `st continue` runs `git rebase --continue`, but git wants `git rebase --skip` for an empty commit — so the user was genuinely stuck with no working stax path forward.

## Fix

Two changes, both in `src/git/repo.rs` at the lowest shared rebase layer so every path (restack, sync, cascade, move, reorder, reparent, fold, upstack, merge-rebase, TUI) benefits:

1. **Prevention** — pass `--empty=drop` to the shared `git rebase` invocations (`rebase_in_path`, `rebase_onto_upstream_in_path`). Commits that become empty are dropped automatically, which is the correct semantic when restacking a stack after a parent has merged. New restacks never hit the spurious stop.
2. **Recovery** — `rebase_continue` (behind `st continue`) now detects a rebase that is stopped-but-conflict-free and auto-runs `git rebase --skip`, looping until the rebase completes or a real conflict appears. This unblocks anyone **already** stuck in this state.

`--empty=drop` only affects commits that *become* empty during the rebase; commits that *start* empty are governed by `--keep-empty` (default keep), so intentional empty commits are preserved. Real file conflicts still stop and report exactly as before.

## Verification

- Added two regression tests in `tests/conflict_handling_tests.rs`:
  - `test_restack_drops_empty_commit_instead_of_stopping` — a two-commit branch where the first commit becomes empty after trunk advances (two commits are required to bypass the whole-branch-integrated squash-merge short-circuit); restack now succeeds and drops the empty commit.
  - `test_continue_auto_skips_empty_commit` — forces the stuck state with `git rebase --empty=stop`, then confirms `st continue` auto-skips and finishes.
- `cargo fmt --check`, `cargo check`, clippy clean on changed files.
- Full `conflict_handling_tests::` + `continue_tests::` suites green (27 tests); both new tests pass.
- Full native suite run: only pre-existing, unrelated `cli_tests::test_shell_setup_*` failures (verified identical on `main`, from #727).

## Risk

Low. The change is additive at the git-invocation layer and does not alter conflict handling, metadata, CLI surface, or the transaction/undo system.

### DIFF
--- a/src/git/repo.rs
+++ b/src/git/repo.rs
@@ -1161,7 +1161,7 @@ impl GitRepo {
     }
 
     fn rebase_in_path(&self, cwd: &Path, onto: &str) -> Result<RebaseResult> {
-        self.rebase_with_args_in_path(cwd, &["rebase", onto])
+        self.rebase_with_args_in_path(cwd, &["rebase", "--empty=drop", onto])
     }
 
     fn rebase_onto_upstream_in_path(
@@ -1170,7 +1170,7 @@ impl GitRepo {
         onto: &str,
         upstream: &str,
     ) -> Result<RebaseResult> {
-        self.rebase_with_args_in_path(cwd, &["rebase", "--onto", onto, upstream])
+        self.rebase_with_args_in_path(cwd, &["rebase", "--empty=drop", "--onto", onto, upstream])
     }
 
     fn reset_hard_in_path(&self, cwd: &Path, target: &str) -> Result<()> {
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #728** 👈

PR #730, opened 2026-08-03
feat(refresh): add --all-stacks to refresh every stack in one pass

cesarferreira:
## Problem

`st refresh` is hard-scoped to the stack you happen to be standing on. Both the restack scope (`sync.rs`, `unfrozen_restack_scope`) and the submit fetch refs (`refresh.rs`) derive from `stack.current_stack(&current)`.

Two consequences:

- **N stacks means N cycles.** Keeping three independent stacks current is three `st checkout` + `st refresh` rounds, each re-fetching and re-syncing trunk from scratch.
- **From trunk it does almost nothing.** `current_stack` on trunk resolves to empty, so the restack scope is empty and there is no stack to submit — trunk gets synced and that's it. That's the exact moment you'd most want "bring everything up to date."

## Solution

`st refresh --all-stacks` fetches and syncs trunk **once**, then walks every stack root — tracked branches whose parent is trunk, via `stack.children(&trunk)`, the same set `st restack --all` uses — restacking and submitting each in turn.

```
▸ Updating all stacks
  ├─ Sync trunk (once)
  ├─ Restack every stack onto updated parents
  └─ Push branches and update PRs

  1. feat/parser  (2 branches)
     feat/parser #711
     feat/parser-2 (no PR)
  2. fix/flaky-test  (1 branch)
     fix/flaky-test #718
```

The plan prints before anything is touched; `-y`/`--force` skips the confirmation.

## Design notes for reviewers

Four decisions that aren't obvious from the diff:

**1. `--all-stacks`, not `--all`.** `st restack --all` already means *"all branches in the current stack"*. Reusing the name on a sibling command for *"all stacks"* would put two different scopes behind one flag name.

**2. The sync call keeps `restack = true`.** It looks redundant — we restack each stack ourselves right after — but two fail-closed guards in `SyncContext` are gated on that flag: the bail when the fetch failed, and the bail when local trunk didn't reach the fetched remote trunk. Passing `false` would silently disarm both and let every stack restack onto a stale trunk. Sync therefore restacks whichever stack we started on, and the loop re-visits it as a no-op.

**3. `--all-stacks` requires a clean tree (or `--auto-stash-pop`).** Plain `st refresh` tolerates a dirty tree because sync holds its own stash *across* its restack and pops it in `restore_stash`. The per-stack restacks here run *after* that pop, so they'd hit `DirtyWorktree` partway through the loop. A guard up front turns a confusing mid-loop failure into a clear message naming `--auto-stash-pop`.

**4. Submit checks out each stack's root, not its tip.** `SubmitScope::Stack` resolves to `ancestors + self + descendants`. From a root that covers the whole subtree; from a leaf it would silently miss sibling subtrees in a branching stack.

New `restack::run_stack_containing()` exists because the existing restack entry points infer their scope from the current branch. It takes an explicit `RestackScope::StackContaining(root)`. `restack::run`'s signature is untouched — `cascade.rs` calls it positionally.

## Conflict handling

Stops at the first conflict and reports what got done:

```
✗ Stopped at 'feat/parser'
  Done: fix/flaky-test
  Remaining: chore/deps
  Resolve the conflict, run `st continue`, then re-run `st refresh --all-stacks`. Finished stacks are no-ops on the re-run.
```

No persisted cross-stack resume plan — re-running is idempotent, since completed stacks become no-ops. `ConflictStopped` propagates **unwrapped** (the loop's error branch `downcast_ref`s it before attaching any `.context`) so `entrypoint.rs` still maps it to `exit_codes::CONFLICT` and suppresses the duplicate `Error:` line.

Branches checked out in another worktree fail loudly rather than silently — both the restack call and the pre-submit checkout attach context naming the offending stack. That failure mode is far more likely at N stacks than at one.

Frozen branches needed no new code: `restack_inner` already filters `BranchMetadata::is_frozen` out of the scope and reports them in the receipt.

## Verification

- [x] `cargo check`, `cargo fmt --check`
- [x] 2 unit tests in `src/commands/refresh.rs` — `collect_stack_plans` root ordering/descendants, `root_of` resolution
- [x] 5 integration tests in `tests/refresh_all_stacks_tests.rs` (registered in `all_tests.rs`):
  - plan lists every root and branch
  - every stack actually ends up rebased onto the new trunk
  - **plain `st refresh` leaves other stacks untouched** — regression guard for the unchanged path
  - original branch is restored afterwards
  - dirty tree without `--auto-stash-pop` is rejected, with the flag named in the error
- [x] `test_refresh_help` asserts the flag appears in help
- [x] `make test` — **2285 passed, 0 failed**

`git diff` on `refresh.rs` shows exactly one non-additive line in the whole file (the `use anyhow::...` widening); the non-`--all-stacks` path through `run` is otherwise byte-identical.

## Known limitation

`st undo` rolls back one stack at a time in this mode — sync opens its own transaction and each per-stack restack opens an `OpKind::Restack` one. Unifying them into a single transaction is a larger change; deferring it.

## Risk

Low. Entirely behind a new opt-in flag; the existing code path is unchanged and guarded by a regression test.

## Note on CI

The clippy gate will fail on `src/commands/shell_setup.rs:573` (`clippy::if_same_then_else`). That is **pre-existing on `main`** — reproduced at `09fbfb3` with this branch's changes stashed, and it appears to have come in with #727. Nothing in this diff touches that file; it needs its own fix to unblock the gate.

### DIFF
--- a/docs/commands/reference.md
+++ b/docs/commands/reference.md
@@ -41,6 +41,7 @@ st --trace status --json >/dev/null
 | `st refresh` | | Sync trunk without merged-branch cleanup, restack, then push and create/update PRs for the current stack (`st update` is a back-compat alias) |
 | `st refresh --force --yes --no-prompt` | | Run the full refresh flow without sync or submit prompts |
 | `st refresh --verbose` | | Same as `st refresh`, with detailed sync/restack/submit timing |
+| `st refresh --all-stacks` | | Refresh every stack in the repo (fetch/trunk sync once, then restack + submit each stack); stops at the first conflict |
 | `st restack` | | Rebase current stack locally — auto-normalizes missing/merged parents; `--stop-here` limits scope |
 | `st cascade` | | Restack the stack and submit updates, without fetching trunk (offline-friendly) |
 | `st diff` | | Show per-branch diffs vs parent |
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -557,6 +557,9 @@ pub(crate) enum Commands {
         /// Auto-stash and auto-pop dirty target worktrees during sync/restack
         #[arg(long)]
         auto_stash_pop: bool,
+        /// Refresh every stack in the repo, not just the current one
+        #[arg(long)]
+        all_stacks: bool,
     },
 
     /// Checkout a branch in the stack
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -447,6 +447,7 @@ pub fn run() -> Result<()> {
             yes,
             no_prompt,
             auto_stash_pop,
+            all_stacks,
         } => commands::refresh::run(
             no_pr,
             no_submit,
@@ -456,6 +457,7 @@ pub fn run() -> Result<()> {
             yes,
             no_prompt,
             auto_stash_pop,
+            all_stacks,
         ),
         Commands::Checkout {
             branch,
--- a/src/commands/refresh.rs
+++ b/src/commands/refresh.rs
@@ -1,8 +1,11 @@
 use crate::commands;
 use crate::engine::Stack;
+use crate::errors::ConflictStopped;
 use crate::git::{GitRepo, local_branch_exists_in};
-use anyhow::Result;
+use anyhow::{Context, Result, bail};
 use colored::Colorize;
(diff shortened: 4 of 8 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #730** 👈

PR #743, opened 2026-08-09
test: cover mutually-exclusive flag validation across 8 commands

cesarferreira:
## Motivation

A test-gap audit of the suite found it heavily skewed toward happy paths: ~2,280 tests against 466 `bail!` sites, with only 80 negative-path assertions in `tests/`. This is the first of six PRs closing the verified-genuine gaps.

CLI-level guards that reject invalid flag combinations had no test coverage at all. Each is a one-line rejection of user-facing argument validation — exactly the kind of code that rots silently when flags are added or renamed, because nothing fails when the guard stops firing.

New `tests/flag_validation_tests.rs` (11 tests) covering:

- `resolve.rs:35` — `--max-rounds 0`
- `checkout.rs:218` — explicit branch combined with `--trunk` / `--parent`
- `changelog.rs:246` — multiple `--find` queries
- `changelog.rs:303` — empty `--find` query
- `worktree/create.rs:33` — branch name passed together with `--pick`
- `worktree/ai.rs:63` — `--tmux-session` without an explicit lane name
- `standup.rs:162` — `--plain-text` without `--ai`

Three more guards are covered as in-source unit tests, where a pure validation function already exists to target directly: `worktree/shared.rs:822/825/828` via `build_launch_spec`, and `skills.rs:106/125` via `parse_harness_selection`.

**Finding worth a look:** three audited `bail!` sites are unreachable dead code. `create.rs:1474` (`--insert` + `--below`), `create.rs:1477` (`--below` + `--from`), and `standup.rs:165` (`--style` without `--ai`) are all pre-empted by clap-level `conflicts_with` / `requires` attributes at `args.rs:804`, `:807`, and `:1129`. Those `bail!` arms can never execute. The tests here assert on the error text the CLI actually produces (clap's) rather than the dead strings, so they pin real observable behaviour. Removing the dead arms is left as a follow-up, since this PR is deliberately tests-only.

Excluded as already covered: `submit.rs:161/164/177` (tested at `submit.rs:4354`, `:4362`), `worktree/shared.rs:831/834` (`shared.rs:1757`, `:1772`), `generate.rs:68` (`cli_tests.rs:1278`).

No production behavior changes. Full suite green (2342 tests).

### DIFF
--- a/src/commands/skills.rs
+++ b/src/commands/skills.rs
@@ -669,6 +669,20 @@ mod tests {
     }
 
     #[test]
+    fn parse_harness_selection_rejects_empty_spec() {
+        let err = parse_harness_selection("   ").unwrap_err();
+        let msg = format!("{err}");
+        assert!(msg.contains("--skills requires a value"));
+    }
+
+    #[test]
+    fn parse_harness_selection_rejects_comma_only_spec() {
+        let err = parse_harness_selection(",,,").unwrap_err();
+        let msg = format!("{err}");
+        assert!(msg.contains("--skills requires at least one harness id"));
+    }
+
+    #[test]
     fn test_format_update_command_line_shows_flags() {
         assert_eq!(
             format_update_command_line(false, &SkillsUpdateOrigin::AllFlag),
--- a/src/commands/worktree/shared.rs
+++ b/src/commands/worktree/shared.rs
@@ -1784,6 +1784,55 @@ mod tests {
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743** 👈
    * **PR #744**
      * **PR #745**
        * **PR #746**
          * **PR #747**
            * **PR #748**
              * **PR #750**
                * **PR #751**
                  * **PR #752**
                    * **PR #754**
                      * **PR #755**

PR #744, opened 2026-08-09
test: cover dirty-working-tree and main-worktree guard rails

cesarferreira:
## Motivation

Second of six PRs closing verified test-coverage gaps.

The dirty-working-tree checks in `undo`, `redo`, and `restack` are the last line of defence against clobbering uncommitted work, and none of the three had a test. Same for the `worktree remove` guard that prevents removing the main worktree. These are the guards that stop a destructive operation before it starts — a regression in any of them costs a user their working tree.

## Description of changes / notes for reviewers

New `tests/guard_rail_tests.rs` (8 tests) covering:

- `undo.rs:60`, `redo.rs:57`, `restack.rs:151` — "Working tree is dirty. Please stash or commit changes first.", reached via the `--quiet` branch
- `get.rs:445` and `get.rs:450` — "Branch name cannot be empty."
- `get.rs:456` — "Pass a branch name, not a full ref"
- `worktree/remove.rs:203` — "Cannot remove the main worktree.", for both the implicit current-worktree and explicit-by-name invocations

**Finding worth a look:** `worktree/remove.rs:58` is unreachable. `retire_worktree` carries its own identical `bail!`, but `run_with_mode` (`remove.rs:202-204`) checks `worktree.is_main` and bails first on both invocation paths. The other caller, `promote.rs:74`, is likewise guarded upstream by `promote.rs:29-31`, which bails with a different message before `retire_worktree` is reached. Both tests here therefore assert against the reachable site at `:203`, with a comment in the test file recording this rather than mislabeling the coverage.

**Deliberately excluded:** `undo.rs:78` and `redo.rs:75` ("Cannot undo/redo with dirty working tree") are reachable only when the user answers *no* to an interactive `dialoguer::Confirm`. Without a TTY, `interact()` errors rather than returning false, so a non-interactive test cannot reach them.

Also excluded as already covered — the trunk guards in `branch/delete.rs`, `branch/rename.rs`, `detach.rs`, `absorb.rs`, `edit.rs`, `branch/create.rs:1497`; the `checkout.rs:694` cycle detector (tested at `checkout.rs:910`); and all six `worktree/promote.rs` guards (`worktree_promote_tests.rs`).

No production behavior changes.

PR #745, opened 2026-08-09
test: cover GitHub API failure modes (403, 404, 409, unreachable host)

cesarferreira:
## Motivation

Third of six PRs closing verified test-coverage gaps.

The wiremock suite mocked 279 successful responses and only four failures (11×422, 3×500, 1×401). Nothing exercised rate limiting, a missing repository, a conflict, or an unreachable API — so the retry and timeout configuration at `github/client.rs:199-207` was entirely unexercised, and a regression in error surfacing would reach users before it reached CI.

New `tests/github_failure_tests.rs` (4 tests), reusing the harness from `github_list_tests.rs` (config-injected `remote.api_base_url`, `STAX_GITHUB_TOKEN`, GitHub-shaped git remote). Covers 403 with a rate-limit body, 404 for a missing repository, 409 conflict, and connection failure against `http://127.0.0.1:1`. Each test runs in ~0.1s.

**Finding worth a look — this is a real UX bug, not just a coverage gap.** `enrich_api_error` (`client.rs:243-264`) is not reachable from `stax pr list`. It rewrites 404/401 into a token-permissions hint — GitHub returns 404 rather than 403 for private repos an unauthorized token cannot see — but it is only wired into `find_open_pr_by_head` and `list_open_prs_by_head` (`github/pr.rs:421`, `:482`). `list_open_pull_requests` (`client.rs:544-573`) returns `.context("Failed to list pull requests")?` directly.

The user-visible consequence: running `stax pr list` with an expired or under-scoped token prints a bare "Not Found" with no hint to run `stax auth --from-gh`, while the *same token* produces a helpful message elsewhere in the CLI. Verified by hand against a local HTTP server returning each status.

Wiring `enrich_api_error` into the list path is a production change, so it is left as a follow-up. These tests pin the current behaviour so that change is visible when it happens.

Two things to know before extending this file:

- `RetryConfig::Simple(1)` (`client.rs:18`) means a failing endpoint can be requested **twice**, so failure mocks must not assert an exact hit count of one.
- The unreachable-host case uses no mock server at all. `127.0.0.1:1` refuses instantly (~0.07s), whereas a routable-but-dead host would burn a 10s connect timeout, doubled by the retry.

No production behavior changes.

### DIFF
--- a/tests/all_tests.rs
+++ b/tests/all_tests.rs
@@ -76,6 +76,8 @@ mod freeze_tests;
 mod get_tests;
 #[path = "gh_stack_tests.rs"]
 mod gh_stack_tests;
+#[path = "github_failure_tests.rs"]
+mod github_failure_tests;
 #[path = "github_list_tests.rs"]
 mod github_list_tests;
 #[path = "guard_rail_tests.rs"]
--- a/tests/github_failure_tests.rs
+++ b/tests/github_failure_tests.rs
@@ -1,0 +1,128 @@
+use crate::common;
+
+use common::{OutputAssertions, TestRepo};
+use std::fs;
+use std::path::Path;
+use tempfile::TempDir;
+use wiremock::matchers::{method, path};
+use wiremock::{Mock, MockServer, ResponseTemplate};
+
+fn ensure_crypto_provider() {
+    let _ = rustls::crypto::ring::default_provider().install_default();
+}
+
+fn write_test_config(home: &Path, api_base_url: &str) {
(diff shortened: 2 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743**
    * **PR #744**
      * **PR #745** 👈
        * **PR #746**
          * **PR #747**
            * **PR #748**
              * **PR #750**
                * **PR #751**
                  * **PR #752**
                    * **PR #754**
                      * **PR #755**

PR #746, opened 2026-08-09
test: cover stax prev, tmux popup guard, and watch interval selection

cesarferreira:
## Motivation

Fourth of six PRs closing verified test-coverage gaps.

`stax prev` had zero tests, despite `navigation_tests.rs` covering next/up/down/top/bottom. The "go back to where I was" command — the one most likely to be used reflexively — was entirely unverified, including its failure path when the recorded branch has since been deleted.

New `tests/prev_watch_tmux_tests.rs` (8 tests) plus an in-source test module in `watch.rs` (3 tests):

- the `prev` happy path, round-tripping between two checked-out branches
- the no-previous-branch message on a fresh repo
- the previous-equals-current no-op
- the `p` alias
- `navigate.rs:216` — "Previous branch '<x>' no longer exists." after the recorded branch is deleted out from under it
- `tmux.rs:162` — "Not inside a tmux session."
- `watch --help` and clap rejection of a non-numeric `--interval`
- `watch.rs:260` `adaptive_interval`, pinning the 15s/60s/120s polling tiers against the `DEFAULT_INTERVAL_ACTIVE/IDLE/QUIET` constants

**Note on the tmux test —** it removes `TMUX` from the child environment rather than setting it empty. `tmux.rs:161` uses `std::env::var("TMUX").is_err()`, which treats an empty value as *present*, so a naive `TMUX=""` test would pass without ever reaching the guard, and would shell out to a real `tmux display-popup` for anyone running the suite inside tmux. It therefore uses `IsolatedProcessEnv::command()` plus `.env_remove("TMUX")`.

Every asserted string was confirmed against the built binary rather than read off the source — earlier PRs in this stack turned up `bail!` sites whose text never actually reaches the user. Both target sites here are live and their wording matches.

**Deliberately excluded:** `stax watch`'s main loop is an unbounded `loop { … sleep }` with no exit condition, no single-shot mode, and no non-TTY early exit, so only `--help`, argument rejection, and the pure `adaptive_interval` helper are reachable without a production seam. #748 adds that seam. Also excluded: `tmux popup`'s success path (spawns real tmux), and the tmux status-bar formatter, which already has 11 tests at `tmux.rs:180-299`.

No production behavior changes.

### DIFF
--- a/src/commands/watch.rs
+++ b/src/commands/watch.rs
@@ -287,3 +287,98 @@ fn adaptive_interval(ci_statuses: &[BranchCiStatus], stack: &Stack, branches: &[
     // Nothing active — back off
     DEFAULT_INTERVAL_QUIET
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::ci::CheckRunInfo;
+    use crate::engine::stack::StackBranch;
+    use std::collections::HashMap;
+
+    fn empty_stack() -> Stack {
+        Stack {
+            branches: HashMap::new(),
+            trunk: "main".to_string(),
+        }
+    }
+
+    fn ci_status_with_check(branch: &str, status: &str) -> BranchCiStatus {
+        BranchCiStatus {
+            branch: branch.to_string(),
+            sha: "abc123".to_string(),
+            sha_short: "abc123".to_string(),
+            overall_status: None,
+            check_runs: vec![CheckRunInfo {
+                name: "build".to_string(),
+                status: status.to_string(),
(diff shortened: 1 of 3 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743**
    * **PR #744**
      * **PR #745**
        * **PR #746** 👈
          * **PR #747**
            * **PR #748**
              * **PR #750**
                * **PR #751**
                  * **PR #752**
                    * **PR #754**
                      * **PR #755**

PR #747, opened 2026-08-09
test: cover Transaction rollback-on-drop and the unexercised tx API

cesarferreira:
## Motivation

Fifth of six PRs closing verified test-coverage gaps.

`src/ops/tx.rs` is the transaction core every `stax undo` depends on. It already had five unit tests, but they covered only `record_optional_after`, `set_head_branch_after`, incremental `snapshot`, and the two finalization paths. The `Drop` implementation at `tx.rs:368-377` — which marks an unfinished transaction as failed, and is the mechanism that makes an *interrupted* operation recoverable — had no coverage at all.

Extends the existing `#[cfg(all(test, unix))]` module with 10 new tests (15 total):

- drop-without-finish on a snapshotted transaction, asserting the receipt is reloaded **from disk** as `OpStatus::Failed` with "Transaction dropped without finishing"
- drop-without-finish on an un-snapshotted transaction, asserting no receipt is left behind
- the metadata-ref round trip (`plan_metadata_ref` / `record_metadata_ref_after`)
- `record_known_after` and `record_known_metadata_after` recording a deletion as `None`
- the remote-ref round trip (`plan_remote_branch` / `record_remote_after`)
- `record_all_after` across every planned branch
- `snapshot_branch_with_metadata`
- persistence of `set_plan_summary`, `set_auto_stash_pop`, and `push_completed_branch`

Assertions reload the persisted receipt from disk rather than reading in-memory struct fields, so they pin what a subsequent `stax undo` would actually find.

**Note for anyone following `patterns.md`:** its transaction snippet is stale. The real signatures take a repo — `Transaction::begin(kind, repo, quiet)`, `plan_branch(&mut self, repo, branch)`, `record_after(&mut self, repo, branch)`.

**One observation, recorded rather than acted on:** the caveat motivating `record_known_after` / `record_known_metadata_after` — that libgit2's cached refdb may not observe refs written by `git` subprocesses — did not reproduce here. `record_all_after`, `record_metadata_ref_after`, and `plan_remote_branch` all correctly observed ref state written by subprocess `git commit` / `update-ref` through an already-open `GitRepo` handle, consistent with libgit2 re-reading loose refs per lookup. The caveat appears specific to `sync`'s usage pattern rather than general. The explicit-value recording helpers remain the safe choice; nothing was changed.

Tests only. This branch touches no integration-test registration and is independent of the rest of the stack — it could be rebased onto `main` standalone if preferred.

### DIFF
--- a/src/ops/tx.rs
+++ b/src/ops/tx.rs
@@ -610,4 +610,414 @@ mod tests {
         assert_eq!(receipt.summary_status(), &OpStatus::Failed);
         assert_eq!(receipt.error.unwrap().message, "conflict");
     }
+
+    #[test]
+    fn dropping_a_snapshotted_transaction_marks_the_receipt_failed() {
+        let temp = tempfile::tempdir().unwrap();
+        let op_id = "drop-snapshotted".to_string();
+        let receipt = OpReceipt::new(
+            op_id.clone(),
+            OpKind::Restack,
+            temp.path().display().to_string(),
+            "main".into(),
+            "feature".into(),
+        );
+        let transaction = Transaction {
+            receipt,
+            git_dir: temp.path().to_path_buf(),
+            workdir: temp.path().to_path_buf(),
+            snapshotted: true,
+            finished: false,
+            quiet: true,
+            backed_up: 0,
+        };
+
+        drop(transaction);
+
+        let reloaded = OpReceipt::load(temp.path(), &op_id).unwrap();
+        assert_eq!(reloaded.status, OpStatus::Failed);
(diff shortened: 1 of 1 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743**
    * **PR #744**
      * **PR #745**
        * **PR #746**
          * **PR #747** 👈
            * **PR #748**
              * **PR #750**
                * **PR #751**
                  * **PR #752**
                    * **PR #754**
                      * **PR #755**

PR #748, opened 2026-08-09
feat: add stax watch --iterations to bound the refresh loop

cesarferreira:
## Motivation

Last of six PRs closing verified test-coverage gaps, and **the only one that changes production code.**

`stax watch` was the only command in the CLI with no coverage of its actual behaviour, and it was untestable by construction: `watch::run()` is an unbounded `loop { … sleep }` with no exit condition, no single-shot mode, and no non-TTY early exit. #746 could only reach `--help`, argument rejection, and the pure `adaptive_interval` helper.

The flag is useful beyond tests: it makes `stax watch` scriptable — `stax watch --iterations 1` as a one-shot CI snapshot, or bounding a poll inside a wrapper script without having to signal the process.

Adds `--iterations <N>`, which stops the watch loop after N refreshes.

The production diff is four lines of behaviour, no refactoring:

- the clap field on `Commands::Watch` (`args.rs`)
- the dispatch destructure (`cli/mod.rs`)
- `--iterations 0` rejected with "--iterations must be at least 1"
- the bounded-mode early return

The early return sits **after** the table renders but **before** the trailing sleep, so the final iteration exits immediately rather than waiting out a 15–120s interval. That placement is the whole reason the tests finish in ~1s instead of timing out.

Tests cover single-iteration exit, a two-iteration run with `--interval 1` (~1.1s wall clock, verified non-hanging), the zero rejection, and the help text.

**Worth knowing for anyone extending these tests:** `watch::run` resolves `RemoteInfo` and constructs a `ForgeClient` *before* entering the loop, so a test repo needs a GitHub-shaped `origin` and `STAX_GITHUB_TOKEN` set even though no request is ever made — with no tracked branches, `branches_to_watch.is_empty()` short-circuits before `fetch_ci_statuses`. `TestRepo::new_with_remote()` is **not** sufficient here, since it points `origin` at a local bare repo that does not resolve as a GitHub forge.

Full suite green: 2342 tests, 0 failures.

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -938,6 +938,9 @@ pub(crate) enum Commands {
         /// Polling interval in seconds (overrides adaptive default)
         #[arg(long, short)]
         interval: Option<u64>,
+        /// Stop after N refreshes instead of running until interrupted
+        #[arg(long, value_name = "N")]
+        iterations: Option<usize>,
     },
 
     /// tmux integration: status bar string and popup viewer
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -602,7 +602,11 @@ pub fn run() -> Result<()> {
             verbose,
             oneline,
         ),
-        Commands::Watch { current, interval } => commands::watch::run(current, interval),
+        Commands::Watch {
+            current,
+            interval,
+            iterations,
+        } => commands::watch::run(current, interval, iterations),
         Commands::Tmux { command } => commands::tmux::run(command),
         Commands::Split {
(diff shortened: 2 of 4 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743**
    * **PR #744**
      * **PR #745**
        * **PR #746**
          * **PR #747**
            * **PR #748** 👈
              * **PR #750**
                * **PR #751**
                  * **PR #752**
                    * **PR #754**
                      * **PR #755**

PR #752, opened 2026-08-09
test: make the CLI-unreachable guards reachable instead of deleting them

cesarferreira:
## Motivation

Four `bail!` arms are unreachable from the CLI because clap rejects the bad combination first:

- `create.rs:1474` / `:1477` — `--insert`+`--below` and `--below`+`--from`, pre-empted by `conflicts_with_all` in `args.rs`
- `standup.rs:165` — `--style` without `--ai`, pre-empted by `requires = "ai"`
- `remove.rs:58` — "Cannot remove the main worktree", pre-empted by every caller

These were surfaced by #743 and #744. The tempting read is "dead code, delete it." **That is the wrong trade.** Each sits inside a plain function taking already-parsed values — a `bool`, an `Option<String>`, a `&WorktreeInfo` — not inside CLI parsing. The clap attribute protects one entry point; the function guard protects every *other* caller, present and future. Deleting them swaps a real safety property for a cosmetic coverage win.

Two of the sites make that concrete rather than hypothetical:

- `branch::create::run` is dispatched from **three** places in `cli/mod.rs` (`:535`, `:731`, `:834`), backed by three separate clap variants that each repeat `conflicts_with_all` by hand (`args.rs:807`, `:1333`, `:1522`). The invariant currently rests on three hand-copied attributes that no test ties together. A fourth alias added without the attribute is a plausible mistake, and the function guard is what catches it.
- `retire_worktree` is `pub(crate)` and reached from **four call paths across three subsystems**: the CLI (`remove.rs:240`), the sync engine (`sync.rs:3722`), the TUI (`tui/worktree/app.rs:619`), and `promote.rs:74` — the latter three via the `remove_worktree_with_hooks` wrapper. Each has its own separate guard in its own file. The TUI's guard runs in `request_delete` (`app.rs:264`) while the removal happens later on a background thread that **re-resolves the worktree by name**, so the check and the use are not atomic. `WorktreeInfo` is `pub` with `pub` fields. This guard is the one place the invariant is actually enforced.

So the fix is to make each guard reachable from a test — turning "unreachable dead code" into "covered defensive depth".

**No behavior changes.** Every guard, message, and check order is preserved.

- **`create.rs`** — extracts the two flag checks into `validate_placement_flags`, called as the first line of `resolve_create_placement`. The guards fired before `&GitRepo` was ever touched, so pulling them into a repo-free function makes them directly unit-testable without standing up a real repository inside a `src/` test.
- **`remove.rs`** — extracts the is_main + path-exists checks into `ensure_removable_worktree`, now called from **both** `retire_worktree` and `run_with_mode`. Those two blocks were byte-identical, so this also removes a copy-paste divergence risk. `run_with_mode`'s early call is kept **deliberately**: it produces the error before the pre-remove hook runs, and dropping it would change observable ordering.
- **`standup.rs`** — no source change needed; the bail is the second statement in `run` and returns before any I/O, so the test calls `run` directly.
- Updates the comment in `guard_rail_tests.rs`, which said the `retire_worktree` guard is "unreachable from `stax worktree remove`". True as scoped, but it read as a broader claim than it should — it now names the sync, TUI, and promote paths and points at the new unit tests.

### Verification

57 targeted tests pass (`flag_validation_tests`, `guard_rail_tests`, and the three new unit-test modules). The `sync_` and `worktree_` suites that reach `remove.rs` transitively are green — worth noting because `sync.rs:4782` already has a test that sets `is_main = true`. `make lint` and `make test` (2353 tests) both pass.

### Deliberately not changed

- **None of the four guards are deleted.** I checked whether any qualified under "exactly one provably-guarded caller, private, unlikely to grow callers". None do: `standup::run` is `pub`, `retire_worktree` has four callers, and the `create` guards back three separate clap variants.
- **No clap attributes are touched.** The CLI-level conflicts are already covered by `tests/flag_validation_tests.rs` and stay exactly as they are; this PR adds the second layer of coverage, it does not move the first.
- **No message strings changed** — `tests/flag_validation_tests.rs` and `tests/guard_rail_tests.rs` assert them and stay green.

### DIFF
--- a/src/commands/branch/create.rs
+++ b/src/commands/branch/create.rs
@@ -1463,6 +1463,16 @@ fn create_branch_with_banner(
     Ok(())
 }
 
+fn validate_placement_flags(from: Option<&str>, insert: bool, below: bool) -> Result<()> {
+    if insert && below {
+        bail!("`--insert` and `--below` cannot be used together");
+    }
+    if below && from.is_some() {
+        bail!("`--below` cannot be used with `--from`");
+    }
+    Ok(())
+}
+
 fn resolve_create_placement(
     repo: &GitRepo,
     current: &str,
@@ -1470,12 +1480,7 @@ fn resolve_create_placement(
     insert: bool,
     below: bool,
 ) -> Result<CreatePlacement> {
-    if insert && below {
-        bail!("`--insert` and `--below` cannot be used together");
-    }
-    if below && from.is_some() {
-        bail!("`--below` cannot be used with `--from`");
-    }
+    validate_placement_flags(from.as_deref(), insert, below)?;
 
     if below {
         let meta = resolve_below_current_metadata(repo, current)?;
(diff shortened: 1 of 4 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743**
    * **PR #744**
      * **PR #745**
        * **PR #746**
          * **PR #747**
            * **PR #748**
              * **PR #750**
                * **PR #751**
                  * **PR #752** 👈
                    * **PR #754**
                      * **PR #755**

PR #754, opened 2026-08-09
fix: stop stax ci and stax ready reporting "no CI" when the API call failed

cesarferreira:
## Motivation

`stax ci` on a repository with an expired or under-scoped token prints `no CI` and exits 0. That is not an unhelpful answer — it is a **wrong** one. The command whose entire purpose is to report CI state confidently reports that there is none, and the user gets no signal that anything went wrong.

The swallowing happens in three places, and the one that actually bites is not the one you'd guess:

- `commands/ci.rs:369-372` — `Err(_) => (None, Vec::new())` discards every `fetch_checks` failure. **This is the dominant one**; fixing the layer below it alone would change nothing user-visible.
- `commands/ready.rs:361-362` — `.unwrap_or_default()` on the same call.
- `github/checks.rs:70` — `Err(_) => return Ok((None, Vec::new()))`.

`stax ci` compounds it by `.ok()`-ing the PR lookups first (`ci.rs:349`, `:358`), so a dead token produces a fully-populated, fully-fabricated "everything is fine, there's just no CI here".

**The `--watch` case is worse.** `all_checks_complete` (`ci.rs:1151-1162`) treats an empty `check_runs` list as *done*. So one transient 500 during `stax ci --watch` made the loop print the success banner and exit 0 — a long-running watch declaring victory because a request failed.

Found while auditing the client for #751, flagged there as deserving its own issue. This is that issue.

**The design question is "is this intentional graceful degradation?", and the answer is no — there is nothing to degrade.** Every caller of `fetch_checks` wants the error:

| Caller | Command | CI is… |
|---|---|---|
| `commands/ci.rs:368` | `stax ci` | the point |
| `commands/ready.rs:360` | `stax ready` CLI + TUI | the point |
| `application/ci.rs:66` `load_ci` | main TUI pane | decoration — **already propagates** |

The commands that show CI *opportunistically* — `stax status` (`status.rs:164`), `stax log` (`log.rs:144`), `stax tmux` (`tmux.rs:111`) — read the offline `CiCache` and never make a network call. `stax watch` does fetch, but already falls back to cache on `Err` (`watch.rs:60-63`), so it keeps degrading exactly as today.

And the one decorative caller already does the right thing: `load_ci` maps errors through `provider_error()` into `CiUpdate::Unavailable`. **So this design is not new** — it's the pattern already shipped on the only caller that wanted leniency. This makes the other two consistent with it. GitLab (`gitlab.rs:513`) and Gitea (`gitea.rs:430`) already propagate; GitHub was the sole outlier.

**"No CI configured" is fully distinguishable from "auth failed", and the code already proved it.** `GET /commits/{sha}/statuses` returns **HTTP 200 with `[]`** when there are genuinely no statuses — it does not 404 — and that case is handled on the success path at `checks.rs:73-75`. `check-runs` likewise returns 200 with `total_count: 0`. The `Err(_)` arm was therefore *never* reachable by the no-CI case; it could only ever turn a real failure into a fake one.

### Changes

- `checks.rs` — both endpoints use the `.context(...)`-then-`enrich_api_error` idiom from #751. Ordering is load-bearing: `enrich_api_error` takes `anyhow::Error` by value.
- `commands/ci.rs` — propagates with the branch name in context. Required turning the `buffer_unordered` stream's item type into `Result<BranchCiStatus>` and collecting through `Result<Vec<_>>` before the existing sort. **The only non-mechanical hunk in the diff.**
- `commands/ready.rs` — one line. `fetch_readiness_rows` already did `match result?`, and the ready TUI already renders per-branch `Err` as `Unavailable`.
- `client.rs::combined_status_state` — drops its `.ok()` / `.ok().flatten()`.

**Five pre-existing tests were relying on the swallow** to tolerate incomplete mocks — they never mocked the commit-statuses endpoint at all, and the swallowed error hid it. They now mock it properly with the 200-plus-`[]` that a real no-CI repo returns. **No assertion was weakened or removed.**

### Behaviour change to review deliberately

`stax ci --watch` now aborts on a fetch failure instead of continuing. I believe that's correct given the false-success bug above, but it's the one place where "we used to keep going" becomes "we stop" — worth a second opinion.

### What I deliberately did not change

- **`combined_status_state` (`client.rs:267`) is dead code** — zero callers repo-wide, so its `.ok()`s were never producing a wrong answer for anyone. Fixed rather than deleted, because deleting orphans `get_check_runs_status` into a test-only fn, trips `dead_code` under `-D warnings`, and would drop 15 unit tests. Separate decision.
- **The `.ok()`s on the PR lookups (`ci.rs:349-358`).** Those are *real* leniency — `stax ci` can legitimately show check state for a branch whose PR metadata is unavailable — and unlike the CI fetch they aren't what the command is about.
- **`all_checks_complete`'s empty-is-done rule.** Correct for a branch that genuinely has no CI; it was only dangerous because failures were being laundered into that state.
- **`--strict` (`ci.rs:1172`)** is unrelated — it means "exit on CI *failure*", not "exit on API error".

Verified: 15 targeted tests, `make lint` clean, `make test` 2358/2358.

### DIFF
--- a/src/application/ci.rs
+++ b/src/application/ci.rs
@@ -161,7 +161,7 @@ mod tests {
     use std::thread;
     use std::time::Duration;
     use tempfile::TempDir;
-    use wiremock::matchers::method;
+    use wiremock::matchers::{method, path_regex};
     use wiremock::{Mock, MockServer, ResponseTemplate};
 
     static CONFIG_ENV_LOCK: Mutex<()> = Mutex::new(());
@@ -707,10 +707,17 @@ mod tests {
         let server = runtime.block_on(MockServer::start());
         runtime.block_on(
             Mock::given(method("GET"))
+                .and(path_regex(r"check-runs$"))
                 .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({
                     "total_count": 0,
                     "check_runs": []
                 })))
+                .mount(&server),
+        );
+        runtime.block_on(
+            Mock::given(method("GET"))
+                .and(path_regex(r"statuses$"))
+                .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!([])))
                 .mount(&server),
         );
 
--- a/src/commands/ci.rs
+++ b/src/commands/ci.rs
@@ -7,7 +7,7 @@ use crate::git::GitRepo;
 use crate::github::pr::CiStatus;
 use crate::notifications::{self, BuiltInSound, Sound};
 use crate::remote::RemoteInfo;
-use anyhow::Result;
+use anyhow::{Context, Result};
 use chrono::{DateTime, Utc};
 use colored::Colorize;
 use futures_util::{StreamExt, stream};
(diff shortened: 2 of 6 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743**
    * **PR #744**
      * **PR #745**
        * **PR #746**
          * **PR #747**
            * **PR #748**
              * **PR #750**
                * **PR #751**
                  * **PR #752**
                    * **PR #754** 👈
                      * **PR #755**

PR #755, opened 2026-08-09
chore: close the gaps in forge API call accounting

cesarferreira:
## Motivation

`record_api_call` feeds the `forge.api.total` and per-operation counters printed by `stax submit --verbose` (`submit.rs:3602`). It was applied opportunistically as methods were added, so roughly half the client's network calls were invisible to it. A user reading that diagnostic to understand why a submit was slow, or to reason about rate-limit headroom, got a number that could be off by a factor of two or more.

**This is a cosmetic undercount in a diagnostic, not a correctness bug** — nothing branches on these counters. The PR is scoped accordingly: add the missing calls, fix one placement bug, stop.

I audited every method in `client.rs`, `pr.rs`, `checks.rs` and `gh_stack.rs` that issues a network request. The gaps:

- **`pr.rs`** — `merge_pr`, `is_pr_merged`, `list_issue_comments`, `list_review_comments`, `get_pr_reviews`, `get_pr_node_id`, `enqueue_pr`
- **`client.rs`** — `get_current_user`, `get_recent_merged_prs`, `get_recent_opened_prs`, `get_reviews_received`, `get_user_open_prs`, `get_check_runs_status`
- **`checks.rs`** — `fetch_commit_statuses`, `fetch_check_runs`

**Two cases where one method makes N requests but recorded 0 or 1:**

- `get_reviews_received` and `get_user_open_prs` each do a search *plus a request per result in a `for` loop*. These now record inside the loop, following the precedent set by `list_open_prs_by_head` (`pr.rs:468`) — the only existing site that got this right.
- `enqueue_pr` makes **two** GraphQL requests: a node-ID lookup and the mutation. Recording `get_pr_node_id` separately fixes this and, as a side effect, also fixes `set_pr_draft`, which shared the same hidden node-ID request.

**One placement bug fixed:** `list_open_issues` recorded once *before* its pagination loop, undercounting by one per extra page. Moved inside the loop.

`checks.rs` needed no refactor to participate — `record_api_call` is `pub(crate)` and `checks.rs` is an `impl GitHubClient` block in the same crate that already reaches `self.octocrab`.

Naming follows the dominant `resource.action` convention with a `graphql.` prefix for GraphQL, matching the existing `graphql.pr_merge_status`.

**Semantics preserved:** `record_api_call` fires *before* the request, so it counts attempts, not successes. That's the existing behaviour at all 22 current sites and is the right one — a failed request still consumes rate limit.

**Tests:** three wiremock unit tests alongside the four that already assert on `api_call_stats`. The counters are readable in-crate via `api_call_stats()` plus the `#[cfg(test)]` `with_octocrab` constructor, so no submit run and no new harness. The two that earn their keep: `test_enqueue_pr_records_node_id_and_mutation` pins the two-requests-one-method case, and `test_list_open_issues_counts_every_page` pins the loop-placement fix. **No new test files.**

**One honest caveat:** instrumenting `get_check_runs_status` is currently *inert*. It makes a real network call, but its only in-crate caller is `combined_status_state`, which is dead code (zero callers repo-wide — see the note in #754). The call is correct and will start counting if that code is ever wired up; it just does nothing today.

### What I deliberately did not change

- **`gh_stack.rs` is scoped out.** It makes zero direct network calls — everything goes through `Command::new("gh")` / `Command::new("git")` subprocesses. Those do hit GitHub, but they're free functions with no `&self` and no tracker access. Counting them needs a different mechanism, and mixing subprocess calls into a counter labelled `forge.api.total` would make the number *less* meaningful, not more.
- **Pure delegators do not record** — `get_pr_review_decision`, `list_all_comments`, `fetch_checks`, and the shared `graphql_data` transport. Recording in any would double-count. `graphql_data` looks like the obvious single choke point, but instrumenting it would double every GraphQL caller that already records.
- **`combined_status_state`** (dead code) and **`get_reviews_given`** (a stub with no network call) are left alone.
- **The legacy inconsistent names stay** — `pulls.convertToDraft` and `pulls.markReadyForReview` are camelCase and `pulls.update-branch` is hyphenated, against the dominant convention. Renaming would change user-visible diagnostic output for no benefit.
- **No test on the printed output** at `submit.rs:3602`. Asserting on formatted diagnostic text is brittle; the counters themselves are pinned at the source.

Checked for stale assertions: the only four `total_requests` assertions are the pre-existing stats tests covering `list_open_prs_by_head`, `find_open_pr_by_head`, `find_pr`, and `get_pr_with_head`. None of those methods were touched, and none asserts on the methods whose totals shift.

Verified: `make lint` clean, `make test` 2361/2361.

### DIFF
--- a/src/github/checks.rs
+++ b/src/github/checks.rs
@@ -59,6 +59,7 @@ impl GitHubClient {
         repo: &GitRepo,
         commit_sha: &str,
     ) -> Result<(Option<String>, Vec<CheckRunInfo>)> {
+        self.record_api_call("checks.commit_statuses");
         let url = format!(
             "/repos/{}/{}/commits/{}/statuses",
             self.owner, self.repo, commit_sha
@@ -121,6 +122,7 @@ impl GitHubClient {
         repo: &GitRepo,
         commit_sha: &str,
     ) -> Result<(Option<String>, Vec<CheckRunInfo>)> {
+        self.record_api_call("checks.check_runs");
         let url = format!(
             "/repos/{}/{}/commits/{}/check-runs",
             self.owner, self.repo, commit_sha
--- a/src/github/client.rs
+++ b/src/github/client.rs
@@ -286,6 +286,7 @@ impl GitHubClient {
 
     /// Get status from GitHub Actions check runs
     async fn get_check_runs_status(&self, commit_sha: &str) -> Result<Option<String>> {
+        self.record_api_call("checks.check_runs");
         let url = format!(
             "/repos/{}/{}/commits/{}/check-runs",
             self.owner, self.repo, commit_sha
@@ -350,6 +351,7 @@ impl GitHubClient {
 
     /// Get the authenticated user's login name
     pub async fn get_current_user(&self) -> Result<String> {
+        self.record_api_call("users.current");
         let user = self.octocrab.current().user().await?;
         Ok(user.login)
     }
@@ -362,6 +364,7 @@ impl GitHubClient {
     ) -> Result<Vec<PrActivity>> {
         let since = Utc::now() - chrono::Duration::hours(hours);
         // Use search API to find only user's merged PRs - much faster than listing all
+        self.record_api_call("search.issues");
         let url = format!(
             "/search/issues?q=repo:{}/{}+author:{}+is:pr+is:merged&sort=updated&order=desc&per_page=30",
             self.owner, self.repo, username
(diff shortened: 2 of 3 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #743**
    * **PR #744**
      * **PR #745**
        * **PR #746**
          * **PR #747**
            * **PR #748**
              * **PR #750**
                * **PR #751**
                  * **PR #752**
                    * **PR #754**
                      * **PR #755** 👈

PR #767, opened 2026-08-13
test: cover single-quoted skill versions

### LINKED ISSUE #762: bug: correctly parse single-quoted stax_version in skill front matter
## Category
Bug

## Evidence
`src/commands/skills.rs:243` parses YAML-frontmatter versions with:

```rust
rest.trim().trim_matches('"').trim_matches('\\').to_string()
```

It strips double quotes and backslashes but not single quotes. Therefore `stax_version: '1.0.0'` is returned as `'1.0.0'`, while the existing unit test at `src/commands/skills.rs:568-572` expects `Some("1.0.0")`. This makes a valid single-quoted skill appear stale to `stax skills list` and causes `stax skills update` to rewrite it unnecessarily.

A targeted validation was attempted with `timeout 90s cargo nextest run test_extract_yaml_single_quotes`; it hit the required 90-second limit while compiling dependencies, so runtime confirmation is inconclusive. The mismatch is statically deterministic.

## Why it matters
Skill files written or edited with normal YAML single-quote syntax are permanently reported out of date, producing misleading status and unnecessary network/write work.

## Suggested fix / test plan
- Strip both single and double quotes from the parsed `stax_version` scalar (or parse the front matter with a YAML parser).
- Keep/fix `test_extract_yaml_single_quotes` so it passes.
- Add an integration regression that seeds a single-quoted installed skill at `PKG_VERSION` and verifies `skills list` marks it current and `skills update` skips it.

_Found by the nightly Hermes self-improvement scout._

cesarferreira:
## Summary

- verify single-quoted YAML `stax_version` values are treated as current by both `skills update` and `skills list`
- retain the unit regression and clarify supported quoted/unquoted scalar syntax

Closes #762

## Verification

- `cargo nextest run test_extract_yaml_single_quotes`
- `make lint-fast`
- `make lint`
- `make test` (2371 passed)

### DIFF
--- a/src/commands/skills.rs
+++ b/src/commands/skills.rs
@@ -226,7 +226,8 @@ fn format_harness_selection_summary(ids: &[String], origin: &SkillsUpdateOrigin)
     format!("Harnesses ({source}): {id_list}")
 }
 
-/// Parse `<!-- stax-skills-version: X.Y.Z -->` or `stax_version: "X.Y.Z"` from the
+/// Parse `<!-- stax-skills-version: X.Y.Z -->` or a quoted/unquoted
+/// `stax_version: X.Y.Z` YAML scalar from the
 /// first 40 lines of a skill file's content.
 pub fn extract_skills_version(content: &str) -> Option<String> {
     for line in content.lines().take(40) {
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #767** 👈

PR #775, opened 2026-08-18
fix: stop submit stalling silently on temporary restacks

cesarferreira:
## Motivation

Reported by Sean Rennie on 0.104.6: `stax ss` stalls for multiple minutes with no output at all once a stack has more than ~3 branches. The last thing printed is `✓ Fetching from origin... done 4.393s`, then nothing.

The stall is the temporary-restack preparation phase in `submit.rs`, which runs between the fetch timer and the `Planning PR operations...` timer and prints nothing while it works. For every branch it did a full `git worktree add --detach` (whole-tree checkout), a `git rebase --onto`, and a `git worktree remove` (`rm -rf`) — sequentially.

Two things make that blow up on a real stack:

- The temp-worktree condition cascades. A single branch needing a restack forces *every* branch above it through the same worktree cycle, because a child publishing on top of a temp ref must itself publish from a temp ref.
- Empty branches were not exempt, so push-only/no-PR branches each paid for a full worktree too.

In a large repo each cycle is tens of seconds, so a 4-5 branch stack is several minutes of dead air. Sean's screenshot shows exactly this shape: one branch `needs restack` plus two empty branches.

- **Visible progress.** Each rebase now runs under a `LiveTimer` (`Preparing restack 1/2: <branch>...`), so the phase reports itself while it happens rather than only via one summary line printed after it finishes.
- **Empty branches no longer get a worktree.** `empty` is defined in submit as `branch_commit == parent_commit`, so an empty branch's rebased tip is *exactly* its parent's rebased tip. It now inherits the parent's prepared ref. When the parent has no temp ref, `refs/heads/<branch>` is already correct and nothing is prepared at all.
- **One worktree, reused.** `ReusableSubmitWorktree` creates a single detached worktree on first use and `git checkout --force --detach <branch>`es between branches. Checking out between adjacent stack branches only touches the files that actually differ, instead of a full checkout plus `rm -rf` per branch.
- The summary line now counts refs actually created rather than branches publishing from them (an inheriting empty branch is not a new ref).
- `--verbose` additionally prints `N rebase(s) across M worktree(s)`.

Rebase targets are now resolved during a planning pass rather than recomputed mid-loop. That pass is what makes the empty-branch inheritance correct: a branch stacked *above* an empty branch must rebase onto the empty branch's inherited temp ref, not onto the stale `refs/heads/<empty>`. Getting this wrong published unrebased commits, and the correctness test below caught it.

### Verification

New `tests/submit_temp_restack_tests.rs` (4 tests), written against the current code first and confirmed failing before the fix:

- `temp_restack_preparation_reports_per_branch_progress` — the phase is not silent.
- `empty_branches_do_not_get_their_own_temp_worktree` — reproduced the bug as `Prepared 3 temporary restack refs` for a `base → empty → tip` stack; now 2.
- `temp_restack_reuses_a_single_worktree_for_every_branch` — asserts `2 rebase(s) across 1 worktree(s)`.
- `temp_restack_publishes_correctly_rebased_commits` — pins the published SHAs: the empty branch lands on the same commit as its parent, and the tip is genuinely rebased onto the published base.

The tests drive the same code path a real `stax ss` takes. `--no-pr` alone routes to the newer application backend (which has no temp-restack phase), so they pass `--no-template` to stay on the PR-creating path while remaining offline.

`make test`: 2381 passed, 0 failed. `cargo clippy --all-targets -- -D warnings`: clean.

### Risk

Behaviour-preserving by intent; the published commits are pinned by test. The reuse path does mean a `git checkout --force` between branches inside the temporary worktree — that worktree is stax-created and detached, never the user's checkout, and is torn down at the end of the phase as before.

### DIFF
--- a/src/commands/submit.rs
+++ b/src/commands/submit.rs
@@ -26,7 +26,7 @@ use futures_util::future::join_all;
 use serde::Deserialize;
 use std::collections::{BTreeSet, HashMap, HashSet};
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::time::{Duration, Instant};
 
@@ -750,8 +750,14 @@ pub fn run(scope: SubmitScope, options: SubmitOptions) -> Result<()> {
         )?;
     }
 
-    let (publish_sources, _temporary_publish_refs) =
-        prepare_publish_sources_for_submit(&repo, &stack, &branches_to_submit, quiet)?;
+    let (publish_sources, _temporary_publish_refs) = prepare_publish_sources_for_submit(
+        &repo,
+        &stack,
+        &branches_to_submit,
+        &empty_set,
+        quiet,
+        verbose,
+    )?;
 
     // Build plan - determine which PRs need create vs update
     let planning_timer = LiveTimer::maybe_new(!quiet, "Planning PR operations...");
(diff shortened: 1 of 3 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #775** 👈

PR #776, opened 2026-08-18
perf: replay temporary submit restacks in the object database

cesarferreira:
## Motivation

Follow-up to #775. That PR made the temporary-restack phase cheaper (one worktree instead of N, empty branches skipped) but left the fundamental cost in place: it still materialised a working tree to compute a rebase.

It doesn't need one. A temporary restack exists only to produce a commit id to push — nothing ever reads the files. Measured on the WayveCode monorepo (95k tracked files):

```
git worktree add      39.2s
git checkout           5.3s
git worktree remove    7.3s
add+remove cycle      46.5s
```

That ~46s is pure filesystem work for a checkout nobody looks at.

Replay the commits directly in the object database with `git merge-tree --write-tree` + `git commit-tree`. `merge-tree` does exactly the three-way merge a cherry-pick does and writes the resulting tree to the object store; `commit-tree` then builds the commit. No index, no working tree, no checkout. The repo already uses this primitive in `check_rebase_conflicts()` (`src/git/repo.rs:2236`).

The result matches `git rebase --onto` in every field that carries meaning: tree, parent, author name/email/date, and the exact message bytes. The committer *timestamp* differs, because it is "now" for both and they run at different moments — real git has the same property (rebasing twice gives different SHAs). `tests/submit_temp_restack_tests.rs::replayed_commits_are_identical_to_a_real_rebase` pins this field by field against an actual `git rebase` run in the same repository.

It falls back to the existing worktree rebase whenever the replay can't be trusted:

- **conflicts** — so the user still gets git's own conflict output rather than a synthesised message
- **merge commits** in the range — `git rebase` flattens or drops these, and a per-commit three-way merge would quietly do something different
- **root commits** — no parent to use as the merge base
- **`commit.gpgsign`** — signing has to be done by git; better slow than silently unsigned

Commits that replay to nothing are dropped, matching rebase's default. That's tested against a real rebase rather than assumed.

The commit message is read with `git log --format=%B -z` and kept as raw bytes. Both details matter: `%B` without `-z` appends its own newline that `commit-tree` would store verbatim, silently producing messages one byte longer than git's; and going through a lossy UTF-8 conversion would corrupt a non-UTF-8 message. Author identity must travel through environment variables, so if it is not valid UTF-8 the replay bails to the worktree path rather than mangling it.

### Measured

Synthetic 5-branch stale stack (3 needing rebase, 2 empty):

| tracked files | 0.104.7 | this PR |
|---|---|---|
| 4,000 | 2.1s | 0.9s |
| 30,000 | 11.5s | 1.8s |

The old path scales with working-tree size; this doesn't. Per-branch, the restack step goes from ~1.1s to ~0.09s and stays flat. On the 95k-file monorepo the replay itself is ~0.27s for 3 commits, against ~46s for one worktree cycle.

### Verification

`tests/submit_temp_restack_tests.rs` grew to 6 tests. The two that matter most for this change:

- `temp_restack_publishes_correctly_rebased_commits` (from #775, unchanged) pins the actual published SHAs — the empty branch lands on its parent's commit, the tip is genuinely rebased onto the published base. This is what proves the replay is correct, and it passed without modification.
- `conflicting_temp_restack_falls_back_to_a_real_worktree` drives a genuine conflict and asserts git's own `CONFLICT` output still reaches the user.

`make test`: 2383 passed, 0 failed. `cargo clippy --all-targets -- -D warnings`: clean.

### Risk

The fallback list is the main thing to review — I chose to bail rather than guess in each case, so the worst outcome for an unanticipated situation is the old (slow, correct) behaviour rather than a wrong commit. If you think any of the four bail conditions is too conservative, that's a cheap follow-up.

Note this does write loose objects for the replayed commits, same as the worktree path did; they become unreachable if submit later fails and get collected by `gc` as before.

### DIFF
--- a/src/commands/submit.rs
+++ b/src/commands/submit.rs
@@ -3035,7 +3035,7 @@ fn prepare_publish_sources_for_submit(
         );
     }
 
-    let worktrees_created = worktree.finish();
+    let (worktrees_created, replayed) = worktree.finish();
 
     // Empty branches inherit their parent's freshly-created ref. Processed in
     // stack order, so a chain of empty branches inherits transitively.
@@ -3064,9 +3064,10 @@ fn prepare_publish_sources_for_submit(
         );
         if verbose {
             println!(
-                "    {} {} rebase(s) across {} worktree(s)",
+                "    {} {} rebase(s): {} replayed in-memory, {} worktree(s) created",
                 "▸".dimmed(),
                 total,
+                replayed,
                 worktrees_created
             );
         }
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #776** 👈

PR #784, opened 2026-08-23
Improve error outside Git repositories

cesarferreira:
## Summary

- Replace the raw libgit2 error shown outside a Git repository with clear recovery guidance.
- Add an end-to-end regression test that confirms internal libgit2 diagnostics are not shown.
- Use slice fill in the split-hunk toggle path to satisfy the current lint gate without changing behavior.

## Verification

- cargo test --test all_tests cli_tests::repository_commands_outside_git_show_actionable_error -- --exact
- make lint
- make test (2385 tests across 4 binaries)

## Notes

Documentation was not updated because this only improves an undocumented runtime error message.

### DIFF
--- a/src/git/repo.rs
+++ b/src/git/repo.rs
@@ -262,7 +262,11 @@ struct BranchParentMetadata {
 impl GitRepo {
     /// Open the repository at the current directory or any parent
     pub fn open() -> Result<Self> {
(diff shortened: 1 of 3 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #784** 👈

PR #785, opened 2026-08-23
deps: octocrab 0.54.1, adapt to newly-required PR fields

cesarferreira:
## Motivation

`cargo upgrade --incompatible` offers octocrab 0.53.1 → 0.54.1. The branch already carried a broad `Cargo.lock` refresh; this makes the tree actually build and pins the behaviour the upgrade changes.

octocrab 0.54 makes `id`, `number`, `url`, `head`, `base` and `locked` **required** on `PullRequest` — all were `Option` before. A response missing any of them now fails to deserialize rather than producing a struct with holes.

**Adapting to the API change.** The accessors in `github/pr.rs` return values directly instead of `Result`, and `get_user_open_prs` drops its per-field `continue` guards. Both are now unreachable by construction.

**A real hazard the upgrade amplifies.** `get_user_open_prs` skips any PR it cannot read. Skipping one odd PR so the rest still track is correct — but if *every* PR is unreadable the empty result rendered as `No open PRs found for user 'x'`, which is indistinguishable from genuinely having none. Going from 3 required fields to 6 makes that more likely. It now errors and names the offending PR; partial skips warn on stderr.

That is not hypothetical: it is exactly how this upgrade broke `branch track --all-prs` in the test suite, and it took a test failure to notice rather than an error message.

**Test fixtures.** The `track_all_prs` pull request mocks omitted `id`, `url` and `locked`. Real GitHub sends all three, so the fixtures were under-specified rather than the code being wrong — but they deserialized to nothing and made the command look like it found no PRs.

**One assertion relaxed, deliberately.** `test_get_pr_with_head_errors_when_response_missing_head` asserted stax's own context string, `"GitHub PR response missing head"`. The contract it protects — a malformed response yields an actionable error naming the field, not a panic or a silently-empty head — still holds; serde now reports it instead of stax. The assertion checks the durable part.

**New coverage.** `get_user_open_prs` had **no tests at all**, despite the Gitea and GitLab equivalents having them, and it is what `branch track --all-prs` runs on. Added three: happy-path field mapping (including a stacked child whose base is its sibling, so a head/base swap is caught), one-unreadable-PR-skips-and-keeps-the-rest, and all-unreadable-errors. The skip test was verified non-vacuous by making the code abort instead of skip and confirming it fails.

**Unrelated, flagged for splitting.** `clap` 4.6.6 came in with the lockfile bump and now renders a lone alias as `[alias: x]` rather than `[aliases: x]`; `test_help` accepted only the plural. Cosmetic clap change, no stax behaviour change.

The first commit on this branch (`fix(test): compare replayed commits without depending on wall clock`) is **independent of the upgrade** and can be split out or landed first. It fixes a flaky test I introduced in #776: it compared the tip commit's parent SHA against a real rebase, and since committer timestamp feeds the commit id, that only held when both runs landed in the same second. It passed in isolation and in CI and failed under full-suite load — a latent flake in released `main`. It now walks both chains and compares tree, author identity and exact message bytes per commit, which is timestamp-independent and stronger than what it replaces.

### Verification

`cargo clippy --all-targets -- -D warnings`: clean.

Full suite natively (Docker was unavailable, so `make test-native` per CLAUDE.md): **2384 passed, 7 failed**. All 7 are pre-existing environmental failures on macOS — TTY emulation via `script`, `gh` extension probing, parallel-run recovery, worktree hooks. I confirmed this by running the whole suite on a clean `origin/main` worktree and diffing the failure sets: baseline is **8** failures, this branch is the same 7 minus the flaky test fixed above. Strictly better, nothing new.

CI runs the suite on Linux where those 7 pass, so it should be green there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -120,7 +120,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -131,7 +131,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1224,7 +1224,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -1993,7 +1993,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #785** 👈
    * **PR #786**

PR #794, opened 2026-08-24
feat(gh-stack): support gh-stack v0.1.0 atomic stack merges

cesarferreira:
## Motivation

[gh-stack v0.1.0](https://github.com/github/gh-stack/releases/tag/v0.1.0) adds `gh stack merge`, which lands a stack — or part of one — in a single all-or-nothing operation, and is explicitly designed to be automation-friendly via `--yes` / `--merge-method`. stax recommended v0.0.8 and merged native stacks PR-by-PR through the GitHub API, retargeting each base to trunk first. That works, but it is not atomic: a failure partway through leaves some PRs merged and the rest stranded.

This teaches stax about v0.1.0 and lets GitHub do the atomic merge when the stack is actually registered as a native GitHub Stack.

- **Recommend v0.1.0.** `RECOMMENDED_GH_STACK_VERSION` is now `(0, 1, 0)`, and `st doctor` explains the upgrade in terms of `gh stack merge`.
- **Token handling is decoupled from the recommendation.** A new `PUBLIC_STACKS_API_VERSION = (0, 0, 8)` constant is what `gh_stack_command` uses to decide whether to strip `GH_TOKEN`/`GITHUB_TOKEN`. Without this split, raising the recommendation would silently start stripping tokens for v0.0.8/v0.0.9 users and break CI-authenticated `gh stack link`. `link_stack_preserves_injected_token_env_vars_for_v0_0_8` is the regression guard. For the same reason, `doctor`'s legacy-OAuth warning is now gated on `is_legacy_auth_version`, so a v0.0.8 install that is merely below the new recommendation no longer gets a false PAT warning.
- **`st merge --stack` delegates to `gh stack merge`** when *all* of: GitHub remote, `native_stack != "off"`, the repo's native-stack feature cache is `Enabled`, and the installed extension actually exposes `gh stack merge`. Otherwise it takes the existing forge-API path unchanged. The decision lives in a pure `stack_merge_route()` function so it is unit-testable, and it runs after scope calculation but before any forge call — non-GitHub and non-native repos spawn zero extra subprocesses.
- **Three outcomes are handled distinctly.** `Merged` continues into the normal reconcile/cleanup/sync flow. `NotStacked` (the tip PR is not in a native stack after all) falls back to the forge-API path, re-applying the multi-PR merge-method guard first. `Queued` (the base branch uses a merge queue, so gh-stack enqueued instead of merging) returns early *before* reconciliation and branch cleanup — treating it as merged would delete branches whose PRs are still open.
- **The multi-PR method guard is scoped to the forge-API route.** `--squash`/`--rebase` are legitimate on the native route because GitHub reconciles the lower PRs itself; that guard only exists because GitHub's indirect-merge detection needs SHA-preserving merges.
- `ExtensionStatus::Outdated` still means "no `gh stack link`" and is untouched — `gh stack merge` is a separate additive capability probe, so `link` keeps working on v0.0.8.
- Docs updated per the repo policy: `README.md`, `skills.md`, `docs/integrations/github-native-stacks.md`, `docs/workflows/merge-and-cascade.md`, `docs/configuration/index.md`, `docs/commands/core.md`.
- No CLI surface change, no new `OpKind`, and GitLab/Gitea merge paths are unaffected.

## Verification

- `cargo check --all-targets`, `cargo fmt --check`, and clippy with `--all-targets --all-features`: clean.
- 84 targeted tests pass, covering version-status/doctor output, the `merge` capability probe, every `StackMergeOutcome` classification, argument construction (`stack merge <pr> --merge-method <m> --yes`), route selection, and five end-to-end tests asserting delegation, both fallbacks, merge-queue routing without local cleanup, and that an unset feature cache never invokes `gh stack merge`.
- Full suite: 2,403 passed. 15 failures were confirmed pre-existing by re-running them with these changes stashed — they are environment-dependent (a `gh` token in the local OS keyring reaching tests that expect unauthenticated behavior) and unrelated to this change.


cesarferreira:
## Stack Links

* `main`
  * **PR #794** 👈

PR #808, opened 2026-08-26
fix(submit): open a PR via a fork when you lack write access

cesarferreira:
## Motivation

Picks up #709 ("feat(submit): open a PR via a fork when you lack write access" by @kcolford), which went stale after CHANGES_REQUESTED review feedback was never addressed. This re-lands that feature (`stax submit --fork` / `remote.auto_fork = true`, falling back to a fork when you lack push access to the upstream) and fixes every outstanding review item.

Base feature (unchanged from #709, cherry-picked): classify a push rejection as a permission denial, reuse or create a pushable fork, push the branch there, and open the PR with head `<fork_owner>:<branch>`.

Fixes applied on top, from the original review:

**Blockers**
- The `--force-with-lease` "lease" on fork pushes was resolved from a live `ls-remote` immediately before the push, so it always matched and never protected anything — a contributor's pre-existing fork branch could be silently overwritten. Now, the first time stax ever publishes a given branch to a given fork remote, it refuses to proceed unless the fork branch's tip is already an ancestor of the local tip, and records what it published in `refs/stax/fork-published/<remote>/<branch>` so later restacks keep force-pushing without a false positive.
- The "existing `fork` remote points elsewhere" conflict test asserted nothing meaningful (it hit a dead endpoint and never reached the real check). Rewrote it to mount a real mock server and assert on the actual conflict message, plus added a new test asserting stax refuses to overwrite an unrelated existing fork branch.

**Should-fix**
- Fork readiness is now polled via `find_pushable_fork` with backoff instead of blindly retrying the push 2-3 times — fork creation is asynchronous and can take longer than a few seconds on large repos.
- The "existing `fork` remote points elsewhere" conflict check now runs before any mutating GitHub call (previously it ran after `create_fork` had already created a real fork on GitHub).
- A configured `remote.fork_remote` name is now honored even when that remote doesn't exist locally yet, instead of silently creating one named `fork`.
- `maintainer_can_modify` is now only requested on fork PRs (head is `owner:branch`); requesting it unconditionally could 422 on org-owned upstream repos.
- The multi-branch guard now checks the full plan count rather than just the branches that needed a push this run, since the fork head-rewrite applies to every PR create in scope.
- Push failure messages in the fork fallback path now include the original push error instead of dropping it.

**Minor**
- Simplified a redundant substring check in the stale-lease classifier.
- The retry-wait message now respects `--quiet` (moot after the polling rewrite, but the replacement message is gated too).

Left untouched per the original review discussion: the permission classifier's handling of "not authorized" (a protected-branch rejection could in theory be misclassified as a permission denial) — that was raised as an open design question, not a directive, and changing it without more product input risks a worse regression.

## Tests

`cargo check`, `cargo clippy -- -D warnings` clean (2 pre-existing, unrelated `clippy::items_after_test_module` findings in `src/web/{routes,templates}.rs` confirmed present on `main`). Full `make test` (2496 tests) green. `submit_fork_fallback_tests` now has 5 tests, all passing, including two new guard tests for the overwrite-protection and pre-create conflict-check fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,7 @@ These rules apply to AI-generated PR titles and bodies from `generate` and `subm
 | `st ss` | Submit the full stack, open/update linked PRs; temporary-publishes branches that need restack |
 | `st submit --plan [--json]` | Preview fetch, push, PR, retarget, metadata, and stack-link actions without changing local or remote state |
 | `st branch submit` | Submit only the current branch; can publish a temporary rebased head when needed |
+| `st branch submit --fork` | On a denied-for-lack-of-write-access push, retry from a GitHub fork of the upstream repo (single branch; set `[remote] auto_fork = true` for always-on) |
 | `st upstack submit` | Submit current branch and descendants, chaining temporary publish heads when needed |
 | `st reviews --stack [--json]` | Stack-wide review/comment inbox, including inline file/line locations on GitHub (`st comments` remains the current-PR view) |
 | `st next` | Move to the next unmerged branch upstack; fork choices are deterministic |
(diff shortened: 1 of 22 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #808** 👈

PR #826, opened 2026-08-28
fix(sync): snapshot branch metadata in restack transaction, harden needs_restack

### LINKED ISSUE #822: `stax undo` doesn't restore branch metadata after an interrupted sync restack — stack permanently reports "nothing to restack"
### Summary

`stax sync`'s restack phase snapshots **branch heads but not branch-metadata refs** into the operation transaction. When a sync-driven restack is interrupted (rebase conflict) and the user runs `stax undo`, the branch heads roll back correctly but `parentBranchRevision` stays at the post-rebase value.

That leaves the stack with metadata claiming it is already based on the new trunk tip while the head is actually still on the old base. Because `needs_restack()` is a bare SHA string comparison, `stax restack` then reports **"Stack is up to date, nothing to restack"** forever, and `stax validate` / `stax fix` also report clean. The stack is silently stuck and there is no user-visible way to find out why.

Version: `stax 0.107.1`

---

### Reproduction

1. Stack of 3 branches on `main`.
2. `main` moves forward on the remote.
3. `stax sync` — pulls trunk, rebases the bottom branch successfully (writing its metadata to the new trunk tip), then hits a rebase conflict on the second branch. Op fails.
4. `stax undo`.
5. `stax sync` / `stax restack` → "nothing to restack", forever. Files introduced by the new trunk commits are absent from every branch in the stack.

---

### Observed state after the undo

```
main tip:                 cc588cf8d   (a merge commit)
eng-9433 head:            bf7da2f7b
merge-base(main, 9433):   badbf3025   ← branch is genuinely 1 behind
eng-9433 metadata:        {"parentBranchName":"main","parentBranchRevision":"cc588cf8d", ...}
```

`git merge-base --is-ancestor cc588cf8d eng-9433` → false. The stored boundary is a lie.

Reflog confirms undo rolled back the head but not the metadata:

```
bf7da2f7b eng-9433@{0}:                                                  ← stax undo
4e1ff2ff9 eng-9433@{1}: rebase (finish): ... onto cc588cf8dee2749a...    ← sync restack
bf7da2f7b eng-9433@{2}: commit: refactor(authz): ...
```

Interestingly `stax ls` gets this right — it computes real ahead/behind from git and shows `1↓`. So `ls` and `restack` disagree, which is the confusing part for the user:

```
◉  ☁️ fabians/eng-9435 1↑
○  ☁️ fabians/eng-9434 1↑
○  ☁️ fabians/eng-9433 1↑ 1↓     ← ls knows it's behind
○  ☁️ main
```

---

### Root cause

The failed sync receipt (`.git/stax/ops/<id>.json`) records only `refs/heads/*` — there are no `@meta` entries at all:

```json
"local_refs": [
  { "branch": "main",     "refname": "refs/heads/main",     "oid_before": "badbf3025", "oid_after": "cc588cf8d" },
  { "branch": "eng-9433", "refname": "refs/heads/eng-9433", "oid_before": "bf7da2f7b", "oid_after": "4e1ff2ff9" },
  { "branch": "eng-9434", "refname": "refs/heads/eng-9434", "oid_before": "94f33f53b", "oid_after": null },
  { "branch": "eng-9435", "refname": "refs/heads/eng-9435", "oid_before": "9fe26b121", "oid_after": null }
],
"error": { "message": "Rebase conflict", "failed_step": "restack", "failed_branch": "fabians/eng-9434" },
"completed_branches": ["fabians/eng-9433"]
```

`src/commands/sync.rs:2219` plans heads only:

```rust
tx.plan_branches(repo, &restack_scope_order)?;
```

…but the loop directly below it writes metadata at `src/commands/sync.rs:2306`:

```rust
updated_meta.write(repo.inner(), branch)?;
```

So the metadata mutation is outside the transaction's snapshot set and `stax undo` has no `oid_before` to restore.

The helper for exactly this already exists — `Transaction::snapshot_branch_with_metadata` (`src/ops/tx.rs:180`) — and the paired `plan_branch` + `plan_metadata_ref` pattern is used correctly elsewhere in the same file (`src/commands/sync.rs:2445-2450`, and `sync.rs:1508` uses the helper). This one call site just doesn't.

**Suggested fix:** have the restack phase plan the metadata ref alongside each branch head before `tx.snapshot()`, and record `record_known_metadata_after` after each successful per-branch metadata write, mirroring `snapshot_deletion`.

---

### Contributing issue: `needs_restack()` trusts metadata over git

`src/engine/metadata.rs:97`:

```rust
pub fn needs_restack(&self, repo: &Repository) -> Result<bool> {
    if self.source_remote.is_some() {
        return Ok(false);
    }
    let parent_ref = repo.find_branch(&self.parent_branch_name, git2::BranchType::Local)?;
    let current_parent_rev = parent_ref.get().peel_to_commit()?.id().to_string();
    Ok(current_parent_rev != self.parent_branch_revision)
}
```

This never verifies that the parent tip is actually reachable from the branch. Adding an ancestry check — roughly "needs restack if `parent_tip` is not an ancestor of `branch`" — would be self-healing and would have made this a non-event regardless of the snapshot bug. As written, one stale SHA silently disables restack for the whole stack above it.

Same check backs `stax validate`, which reports `PASS All branches up to date` in this state, and `stax fix --dry-run`, which reports `No issues found` — so neither of the repair tools can detect or fix it.

---

### Workaround

`stax branch reparent` without `--restack` rewrites `parentBranchRevision` to the true merge-base (`src/commands/branch/reparent.rs:96-110`), which unsticks the stack:

```bash
stax branch reparent -b <bottom-branch> -p main
stax restack -a
```

cesarferreira:
## Summary
- `stax sync`'s restack phase now plans the `refs/branch-metadata/<branch>` ref alongside each branch head *before* the transaction snapshot, and records the metadata after-OID once a branch's rebase succeeds — so `stax undo` correctly restores `parentBranchRevision` after an interrupted sync restack, instead of leaving it stuck at the post-rebase value.
- `BranchMetadata::needs_restack()` now also checks that the recorded parent revision is an ancestor of the branch (not just a bare SHA match), so a stale/incorrect `parentBranchRevision` self-heals instead of permanently masking drift from `stax restack` / `stax validate` / `stax fix`.

Fixes #822

## Test plan
- [x] `cargo nextest run engine::metadata::` — new ancestry-check unit tests pass
- [x] `cargo nextest run sync_undo_tests::` — new regression test `sync_restack_conflict_undo_restores_branch_metadata` (interrupted sync restack + undo restores both head and metadata)
- [x] `cargo nextest run track_all_local_tests::` — confirms no false positives on the inverted-parent case
- [x] `cargo check && cargo clippy -- -D warnings`
- [x] `make test` (full suite)

### DIFF
--- a/src/commands/range_diff.rs
+++ b/src/commands/range_diff.rs
@@ -49,7 +49,7 @@ pub fn run(stack_filter: Option<String>, all: bool) -> Result<()> {
             None => continue,
         };
 
-        let needs_restack = meta.needs_restack(repo.inner()).unwrap_or(false);
+        let needs_restack = meta.needs_restack(repo.inner(), branch).unwrap_or(false);
         if !needs_restack {
             println!(
                 "\n{} {}",
--- a/src/commands/sync.rs
+++ b/src/commands/sync.rs
@@ -2216,7 +2216,12 @@ impl SyncContext {
                     .tx
                     .take()
                     .context("sync transaction was already finished")?;
-                tx.plan_branches(repo, &restack_scope_order)?;
+                for branch in &restack_scope_order {
+                    tx.plan_branch(repo, branch)?;
+                    if branch != &self.stack.trunk {
+                        tx.plan_metadata_ref(repo, branch)?;
+                    }
+                }
                 let restack_count = branches_to_restack.len();
                 let summary = PlanSummary {
                     branches_to_rebase: restack_count,
@@ -2328,6 +2333,11 @@ impl SyncContext {
 
                             // Record after-OID
                             tx.record_after(repo, branch)?;
+                            let meta_after = resolve_ref_oid(
+                                &self.workdir,
+                                &format!("refs/branch-metadata/{}", branch),
+                            );
+                            tx.record_known_metadata_after(branch, meta_after.as_deref());
                             tx.push_completed_branch(branch);
 
                             if self.verbose {
--- a/src/engine/metadata.rs
+++ b/src/engine/metadata.rs
@@ -93,15 +93,56 @@ impl BranchMetadata {
         Ok(Self::read(repo, branch)?.is_some_and(|metadata| metadata.frozen))
     }
 
-    /// Check if the branch needs restacking (parent has moved)
-    pub fn needs_restack(&self, repo: &Repository) -> Result<bool> {
+    /// Check if the branch needs restacking (parent has moved).
+    ///
+    /// The recorded `parent_branch_revision` can be a lie: an interrupted restack may have
+    /// written the new base before being rolled back, leaving a SHA that no longer sits under
+    /// the branch's tip. A bare SHA comparison would then report the branch clean forever, so
+    /// when the SHAs match we additionally verify the recorded base is still reachable from the
+    /// branch head.
+    pub fn needs_restack(&self, repo: &Repository, branch: &str) -> Result<bool> {
         if self.source_remote.is_some() {
             return Ok(false);
         }
 
         let parent_ref = repo.find_branch(&self.parent_branch_name, git2::BranchType::Local)?;
         let current_parent_rev = parent_ref.get().peel_to_commit()?.id().to_string();
-        Ok(current_parent_rev != self.parent_branch_revision)
+        if current_parent_rev != self.parent_branch_revision {
+            return Ok(true);
+        }
+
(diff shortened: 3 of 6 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #826** 👈
    * **PR #827**

PR #827, opened 2026-08-28
test: cover multi-branch undo and cross-phase metadata idempotency

cesarferreira:
## Summary

Stacked on #826 (issue #822). Adds test coverage for two gaps flagged during review of that fix but not yet covered:

- `sync_restack_conflict_undo_restores_metadata_for_every_successful_branch_in_a_three_branch_stack` (`tests/sync_undo_tests.rs`) extends the 2-branch conflict/undo regression test to a 3-branch stack, proving `stax undo` restores commit + metadata for **every** branch that restacked successfully before the top of the stack hit a conflict, not just the bottommost one.
- `plan_metadata_ref_across_two_phases_keeps_the_earliest_before_oid` (`src/ops/tx.rs`) covers the cross-cutting concern noted in the original plan: when two sync phases (e.g. a reparent/deletion phase and the later restack phase) both plan the same branch's metadata ref within one transaction, `add_metadata_ref`'s idempotency must keep the *first* phase's before-OID, so undo rolls all the way back to the true pre-operation state rather than an intermediate one.

Both were verified to actually catch their target regression: I temporarily broke the relevant guarantee (removed the idempotency check / used a 2-branch instead of 3-branch assertion), confirmed the new test failed with the expected error, then restored the fix.

No production code changes — tests only.

## Test plan
- [x] `cargo nextest run sync_undo_tests:: ops::tx::` — all pass, including the two new tests
- [x] `cargo check && cargo clippy -- -D warnings`
- [x] `cargo fmt --check`
- [x] Full suite (`--no-fail-fast`): 2568/2580 passed; the 12 failures are pre-existing and unrelated (only occur when running tests from inside a linked git worktree — confirmed identical on the parent branch before these test additions)

### DIFF
--- a/src/ops/tx.rs
+++ b/src/ops/tx.rs
@@ -717,6 +717,70 @@ mod tests {
     }
 
     #[test]
+    fn plan_metadata_ref_across_two_phases_keeps_the_earliest_before_oid() {
+        // sync runs multiple phases against the same transaction (e.g. a
+        // deletion/reparent phase followed by a restack phase). If both
+        // phases touch the same branch's metadata ref, plan_metadata_ref's
+        // second call must be a no-op that preserves the FIRST phase's
+        // before-OID -- otherwise undo would only roll back to the
+        // intermediate state written by the first phase, not the true
+        // pre-operation state.
+        use std::process::Command;
+
+        let dir = tempfile::tempdir().unwrap();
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #826**
    * **PR #827** 👈

PR #834, opened 2026-08-29
fix(sync): persist ancestry-verified parentBranchRevision after squash-merge cleanup

### LINKED ISSUE #830: sync stamps `parentBranchRevision` with the trunk tip after reparenting a merged parent's child, poisoning the next restack
### Summary

When `stax sync` cleans up a merged parent branch, it rebases the child onto trunk with a **provenance-aware** rebase, then writes the child's `parentBranchRevision` as **`rev_parse(trunk)` — the trunk tip** rather than the boundary that rebase actually used (`src/commands/sync.rs:1523-1524`).

If the child's own commits do not start at the trunk tip, that boundary is now *behind* the child's history. The rebase itself succeeds and the stack looks fine, so nothing surfaces. The damage lands on the **next** sync: restack reads the stale boundary, and `git rebase --onto <new trunk> <stale boundary>` sweeps in the already-merged parent's commits, replaying them onto a trunk that already contains them.

`stax` already has correct logic for exactly this (`reparent_children_for_deletion`, `sync.rs:4152`, carrying the #120 rationale comment). It never runs here, because by the time the parent is deleted the child's recorded parent has already been rewritten to trunk by the path above — so the selector at `sync.rs:4146` (`info.parent == Some(branch)`) does not match it.

Version: `stax 0.107.1`. Confirmed still present on the default branch.

---

### Reproduction

Stack `main → A → B`, GitHub configured to **squash-merge**.

1. Squash-merge `A`'s PR.
2. `stax sync` — trunk advances, `A` is deleted as merged, `B` is rebased onto trunk. **Succeeds.** `B`'s metadata is now `{parentBranchName: "main", parentBranchRevision: <trunk tip>}`, but `B`'s history still contains `A`'s commits.
3. Squash-merge the next PR down so trunk advances again.
4. `stax sync` → rebase conflict, replaying commits that are already upstream.

---

### Observed

Real trace from a 4-branch stack, two syncs 85 seconds apart.

**10:38:27 `sync` — success.** Trunk `cf1afa61f → c439f67da`; `eng-9434` deleted as merged; `eng-9435` and `eng-9436` restacked.

**10:39:52 `sync` — failed, "Rebase conflict".** Trunk `c439f67da → 8207f4636`; `eng-9435` deleted as merged.

`eng-9436`'s metadata after the first sync:

```
$ git cat-file -p refs/branch-metadata/fabains/eng-9436
{"parentBranchName":"main","parentBranchRevision":"c439f67dab...","frozen":false,"prInfo":{"number":2676,...}}
```

`c439f67da` is the trunk tip as of 10:38:27 — but `eng-9436` does not start there:

```
$ git log --oneline c439f67da..3cda98c75
3cda98c75 refactor(authz): reshape the role-definition API for the roles UI   ← eng-9436's own
c4c4f8ee4 fix: after main merge conflicts                                     ← eng-9435's
6547d00f9 feat(authz): assign every capability to a user-facing area          ← eng-9435's
```

So restack ran `git rebase -i --onto 8207f4636 c439f67da fabains/eng-9436` — a 3-commit range whose first two are already upstream inside the squash `8207f4636`.

The correct boundary was available: `eng-9435`'s tip `c4c4f8ee`, which stax recorded as `oid_before` in its own op receipt, and which passes the ancestry guard the good path uses:

```
$ git merge-base --is-ancestor c4c4f8ee 3cda98c75 && echo YES
YES
```

With `c4c4f8ee` as upstream, only `3cda98c75` replays and there is no conflict.

---

### Root cause

Two write sites disagree, and the naive one runs first.

**The naive one — `src/commands/sync.rs:1512-1524`.** The rebase is provenance-aware (it even runs a `git merge-tree --write-tree` squash-merge integration check), but `RebaseResult` is `enum { Success, Conflict }` (`src/git/repo.rs:2538`), so the boundary it resolved is discarded and the caller writes the trunk tip:

```rust
match repo.rebase_branch_onto_with_provenance(
    child,
    &self.stack.trunk,
    branch, // fallback upstream
    false,
) {
    Ok(_) => {
        if let Some(mut metadata) = BranchMetadata::read(repo.inner(), child)? {
            metadata.parent_branch_name = self.stack.trunk.clone();
            metadata.parent_branch_revision = repo.rev_parse(&self.stack.trunk)?;  // ← wrong
            metadata.write(repo.inner(), child)?;
        }
```

**The correct one — `src/commands/sync.rs:4152`,** which already does the right thing and says so:

```rust
// Preserve the old-parent boundary so restack can run `git rebase
// --onto <new> <old>` precisely. Only use the deleted branch's tip
// when it is still in the child's ancestry; otherwise keep the
// recorded revision (see #120).
let old_parent_boundary = doomed_tip
    .clone()
    .filter(|tip| repo.is_ancestor(tip, child).unwrap_or(false))
    .unwrap_or_else(|| child_meta.parent_branch_revision.clone());
```

It is unreachable in this scenario: `children_to_reparent` (`sync.rs:4146`) selects on `info.parent.as_deref() == Some(branch)`, and the child's recorded parent is already `main`.

**One caveat on attribution.** `refs/branch-metadata/*` is an unversioned blob with no reflog, so I cannot *prove* that line 1524 is the write that produced `c439f67da`. The evidence is that the value is exactly trunk's tip immediately after the 10:38:27 op, and that this is the only site stamping `rev_parse(trunk)` — had it been written during the 10:39:52 op it would read `8207f4636`. Strong, but circumstantial.

**Why `--no-reapply-cherry-picks` doesn't rescue it.** `drop_redundant_commits` is present in `.git/rebase-merge/`, so it was enabled. It cannot help, for two independent reasons: detection runs against `<upstream>` (`c439f67da`), which does not contain the squash at all; and the PR was **squash-merged**, so the upstream commit is the *union* of the two local commits and patch-id-matches neither individually. In a squash-merge workflow the recorded boundary is the only thing keeping a stack correct, which is what makes a wrong write unrecoverable rather than self-healing.

---

### Suggested fix

Have `rebase_branch_onto_with_provenance` return the upstream boundary it resolved, and persist **that** at `sync.rs:1524` instead of `rev_parse(trunk)`. Failing that, apply the same rule the good path uses — the deleted/previous parent's tip when it is still an ancestor of the child, else the recorded revision.

A regression test: stack `main → A → B`, squash-merge `A`, sync, then assert `B`'s `parentBranchRevision` is an ancestor of `B` (rather than asserting a specific SHA) — and that a second sync after trunk advances again replays only `B`'s own commits.

Worth considering alongside it: **an invariant check on write.** Every `parentBranchRevision` write could assert `is_ancestor(revision, branch)`; a boundary that is not in the branch's ancestry is never correct, and both this bug and #679 would have been caught at the moment of writing rather than one sync later.

Separately, `refs/branch-metadata/*` is an unversioned blob with no reflog, so a bad write leaves no trail — reconstructing this required matching the stamped OID against trunk's position in the op receipts. Recording metadata before/after in the op JSON (which already tracks branch OIDs) would make this class of bug diagnosable.

---

### Relationship to prior issues

- **#120** — same end state (ghost commits from a squash-merged parent), but that fix (`reparent_children_for_deletion`) is present and correct; this report is that it is **bypassed** when the child's parent was already rewritten to trunk by `sync.rs:1524` in an earlier sync.
- **#679** — also a stale boundary replaying squash-merged commits, but its repair triggers on `parent tip == merge-base`. Here merge-base is `c439f67da` against a tip of `8207f4636` — the branch is genuinely behind, so that condition never holds.
- **#822** — metadata not restored by `undo` after an *interrupted* restack, leaving the boundary *ahead* of reality ("nothing to restack"). Here no undo was involved, the poisoning sync **succeeded**, and the boundary is *behind* reality, so restack fires and over-replays.

cesarferreira:
## Summary

Fixes #830, plus the same bug class found and fixed at three more sites during review.

`parentBranchRevision` is the boundary the *next* restack feeds straight into `git rebase --onto <parent> <boundary>`, so it must always be a real ancestor of the branch it's recorded for. Five write sites across the codebase violated that invariant by stamping "the current tip of some branch" without checking whether the branch had actually moved onto it.

### The two named/related to the original issue (`src/commands/sync.rs`)

1. **`cleanup_merged_branches`** — the rebase result match arm was `Ok(_)`, so a genuinely **conflicting** rebase (`RebaseResult::Conflict`) was treated identically to success: sync would report `✓ Rebased <child> onto main`, stamp the trunk tip as the boundary, and delete the merged parent branch, even though the rebase never completed. Wrong but nothing fails immediately — it only surfaces on the *next* sync, which replays the merged parent's already-upstream commits and conflicts (worse with squash-merges, where patch-id detection can't recognize the duplicate).
2. **`apply_live_pr_state`** (found during review) — reconciles a branch's recorded parent with GitHub's live PR base (e.g. `gh pr edit --base`, or GitHub auto-retargeting a child PR when an intermediate branch is deleted). Same defect, but with **no rebase attempt at all** — a stronger match for the issue's own real-world "Observed" trace, where a grandchild's metadata parent flipped straight to `main` with no corresponding rebase.

### Three more found during a full-codebase audit of every `parentBranchRevision` write site

3. **`branch/create.rs` `apply_insert_reparenting`** (`stax branch create --insert`) — reparents pre-existing children onto the newly inserted branch's tip with zero rebase. If a child was already stale relative to the parent before the insert ran, that tip was never in the child's real ancestry.
4. **`split.rs`** (`stax split --file`) — the original branch's tip is amended in place (extracted files removed), never rebased onto the new split branch. This one is **unconditional**: it reproduces on every `stax split` invocation, not just a stale-precondition edge case.
5. **`tui/split_hunk/app.rs` `finalize()`** — the hunk-split TUI's equivalent reparent of pre-existing stacked children, same no-rebase shape as plain split.

Sites 3–5 are lower severity than 1–2: `needs_restack()` (`engine/metadata.rs`) has its own independent ancestry-reachability check (not just SHA equality), so `stax log`/`status` still correctly report "needs restack" and the next restack self-heals — this works only because none of these three commands rewrite/squash shared history, so plain git range math still isolates the right commits. `stax submit` bypasses that safety net entirely (it reads `parentBranchRevision` directly as its own rebase-range upstream), so the corrupted window isn't fully inert.

Two narrower cases were found and *not* fixed (only trigger if the `git merge-base` call itself errors, essentially unreachable in practice): `tui/mod.rs:876`, `detach.rs:70/87`. `sweep.rs` independently re-implements the same correct guard already used by `reparent_children_for_deletion` — not a bug, just duplication, worth centralizing separately.

## Fix

- Centralized the ancestry-guard logic (try each candidate boundary in order, keep it only if `is_ancestor(candidate, branch)`, else preserve the previously recorded revision) into `GitRepo::resolve_child_parent_boundary`, used by all five write sites.
- Split `cleanup_merged_branches`'s `Ok(_)` match arm into `Ok(RebaseResult::Success)` and a new `Ok(RebaseResult::Conflict)` arm that stops sync with a proper conflict error (mirroring `restack_phase`) instead of silently continuing and deleting the merged branch out from under an incomplete rebase.

## Test plan

- [x] `test_sync_child_boundary_stays_in_ancestry_when_squash_parent_rebase_conflicts` (site 1) — confirmed failing pre-fix, passing post-fix.
- [x] `test_sync_squash_merged_child_boundary_matches_new_base` — clean-path companion.
- [x] `test_sync_pr_base_reconcile_does_not_poison_boundary_when_trunk_has_advanced` (site 2) — confirmed failing pre-fix (ghost-commit count goes from 1 to 2), passing post-fix.
- [x] `test_create_insert_does_not_poison_stale_childs_boundary` (site 3) — confirmed failing pre-fix, passing post-fix.
- [x] `test_split_file_records_ancestry_valid_boundary_for_original_branch` (site 4) — confirmed failing pre-fix, passing post-fix.
- [x] `finalize_does_not_poison_stale_childs_boundary` (site 5, direct unit test since it isn't reachable through the CLI) — confirmed failing pre-fix, passing post-fix.
- [x] Targeted regression suites (squash-merge, undo, json, conflict handling, continue, restack provenance, PR tracking/base reconciliation, insert, split, hunk-split): 149 tests — pass
- [x] `make test` (full suite, 2586 tests) — pass

### DIFF
--- a/src/commands/branch/create.rs
+++ b/src/commands/branch/create.rs
@@ -1541,9 +1541,19 @@ fn apply_insert_reparenting(repo: &GitRepo, parent_branch: &str, new_branch: &st
     let new_parent_rev = repo.branch_commit(new_branch)?;
     for child in &children {
         if let Some(child_meta) = BranchMetadata::read(repo.inner(), child)? {
+            // `new_branch` is created at `parent_branch`'s tip at the time of insert,
+            // not `child`'s -- if `child` was already stale relative to `parent_branch`,
+            // `new_parent_rev` isn't actually in `child`'s ancestry. Verify before
+            // trusting it, else keep the previously recorded (still-valid) boundary
+            // (see #830); `stax restack --all` (prompted below) then rebases for real.
+            let parent_branch_revision = repo.resolve_child_parent_boundary(
+                child,
+                &[Some(new_parent_rev.as_str())],
+                &child_meta.parent_branch_revision,
+            );
             let updated = BranchMetadata {
                 parent_branch_name: new_branch.to_string(),
-                parent_branch_revision: new_parent_rev.clone(),
+                parent_branch_revision,
                 ..child_meta
             };
             updated.write(repo.inner(), child)?;
--- a/src/commands/split.rs
+++ b/src/commands/split.rs
@@ -205,8 +205,18 @@ fn split_by_file(
     // 10. Update current branch metadata: parent is now the new branch
     let new_branch_rev = repo.branch_commit(&new_branch)?;
     if let Some(mut meta) = BranchMetadata::read(repo.inner(), current)? {
+        // `current`'s tip was amended in place, not rebased onto `new_branch` --
+        // its underlying commit chain below the tip is unchanged, so
+        // `new_branch`'s commit generally isn't in `current`'s real ancestry.
+        // Verify before trusting it, else keep the previously recorded (still-
+        // valid) boundary (see #830); `stax restack` (prompted below) then
+        // rebases for real.
         meta.parent_branch_name = new_branch.clone();
-        meta.parent_branch_revision = new_branch_rev;
+        meta.parent_branch_revision = repo.resolve_child_parent_boundary(
+            current,
+            &[Some(new_branch_rev.as_str())],
+            &meta.parent_branch_revision,
+        );
         try_or_rollback!(meta.write(repo.inner(), current));
     }
 
--- a/src/commands/sync.rs
+++ b/src/commands/sync.rs
@@ -1490,6 +1490,7 @@ impl SyncContext {
                                 );
                             }
 
+                            let mut rebased_children: Vec<String> = Vec::new();
                             for child in &children {
                                 if BranchMetadata::is_frozen(repo.inner(), child)? {
                                     if !self.quiet {
@@ -1508,6 +1509,10 @@ impl SyncContext {
                                     tx.snapshot_branch_with_metadata(&repo, child)?;
                                 }
 
+                                // Tip of the merged parent BEFORE deletion — the boundary
+                                // fallback when the child did not actually move onto trunk.
+                                let merged_parent_tip = repo.branch_commit(branch).ok();
+
                                 // Use existing provenance-aware rebase
                                 match repo.rebase_branch_onto_with_provenance(
                                     child,
@@ -1515,14 +1520,21 @@ impl SyncContext {
                                     branch, // fallback upstream
                                     false,  // auto_stash_pop
                                 ) {
-                                    Ok(_) => {
+                                    Ok(RebaseResult::Success) => {
                                         // Update child's parent metadata to trunk
+                                        let trunk_tip = repo.rev_parse(&self.stack.trunk)?;
                                         if let Some(mut metadata) =
                                             BranchMetadata::read(repo.inner(), child)?
                                         {
                                             metadata.parent_branch_name = self.stack.trunk.clone();
                                             metadata.parent_branch_revision =
-                                                repo.rev_parse(&self.stack.trunk)?;
+                                                resolve_child_parent_boundary(
+                                                    &repo,
+                                                    child,
+                                                    Some(trunk_tip.as_str()),
+                                                    merged_parent_tip.as_deref(),
+                                                    &metadata.parent_branch_revision,
+                                                );
                                             metadata.write(repo.inner(), child)?;
                                         }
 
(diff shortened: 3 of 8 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #834** 👈

Issue #835, opened 2026-08-30
parentBranchRevision writes lack op-receipt tracking at 4 of 5 sites fixed in #830


cesarferreira:
## Context

#830 (fixed in #834) noted that `refs/branch-metadata/*` is an unversioned blob with no reflog, so a bad write leaves no trail — diagnosing the original report required matching a stamped OID against trunk's position in the op receipts. The issue suggested: "Recording metadata before/after in the op JSON (which already tracks branch OIDs) would make this class of bug diagnosable."

That infrastructure already exists (`Transaction::snapshot_branch_with_metadata` / `plan_metadata_ref` / `record_metadata_ref_after`, added in #826) and is used by one of the five `parentBranchRevision` write sites fixed in #834. The other four have no such tracking.

## Current coverage

**Tracked (before/after metadata ref OID recorded in the op receipt):**
- `src/commands/sync.rs` `cleanup_merged_branches` — calls `tx.snapshot_branch_with_metadata(&repo, child)` before mutating and records the after-OID. Fully diagnosable/undoable.

**Not tracked:**
- `src/commands/sync.rs` `apply_live_pr_state` — runs inside `refresh_pr_states`, *before* any `Transaction::begin()` in the sync pipeline. No receipt exists at the point this write happens.
- `src/commands/branch/create.rs` `apply_insert_reparenting` (`stax branch create --insert`) — no `Transaction` anywhere in this file.
- `src/commands/split.rs` (`stax split --file`) — no `Transaction`; uses a bespoke `rollback_split_by_file` closure instead of the receipt system.
- `src/tui/split_hunk/app.rs` `finalize()` — *does* use a `Transaction` (`OpKind::Split`), but `tx.plan_branches(&repo, &affected)` only covers `created_branches` + `original_branch`, never `self.children` (the pre-existing stacked descendants whose metadata the #830-class bug touches). Even for branches it does track, it only calls `plan_branches` (branch ref OID), never `plan_metadata_ref` (metadata ref OID).

## Suggested fix

For each of the four sites above, wire in the same pattern already used by `cleanup_merged_branches`:
- Before the write: `tx.plan_metadata_ref(repo, branch)` (or `snapshot_branch_with_metadata` if the branch ref itself also needs tracking).
- After the write: `tx.record_metadata_ref_after(repo, branch)`.

[...]

PR #840, opened 2026-08-30
fix(create): wrap stax create in a Transaction, closing the last of #830's write sites

### LINKED ISSUE #839: stax create has no Transaction/op-receipt support, so --insert reparenting is undiagnosable and unable to undo
## Context

#835 (fixed in #838) wired op-receipt before/after metadata-ref tracking into 3 of 4 previously-untracked `parentBranchRevision` write sites from #830/#834. The 4th, `src/commands/branch/create.rs` `apply_insert_reparenting` (`stax branch create --insert`), was deliberately deferred rather than folded into #838.

## Why it's deferred rather than a simple wiring job

Every other site fixed in #830/#834/#835 already had a `Transaction` in scope, or one whose scope cleanly matched the mutation being tracked. `branch/create.rs` has neither:

1. **No transaction exists anywhere in the file**, and the honest undo unit is the whole `create` operation — branch creation, the metadata write, `--insert`/`--below` reparenting, and (in the commit-first flow) the rewind of `current`'s ref if something fails partway through. Scoping a `Transaction` to *just* `apply_insert_reparenting` would publish a receipt that `stax undo` targets, but reversing it would only restore the children's metadata — leaving the newly created branch and its commit behind. That's strictly worse than today's "no receipt at all."
2. **Two independent flows, ~10 manual rollback call sites.** `create_branch_with_banner` and `run_commit_first`, plus the `CreateBelowAutoStash` state machine that restores stashes on every one of them. Introducing a transaction means reconciling `Transaction::drop`'s failed-receipt semantics with each of those hand-rolled restores, not just adding a call.
3. **Largest blast radius in the test suite** of any of the touched files — `tests/create_rollback_tests.rs` (22K), `tests/create_insert_tests.rs`, `tests/create_below_tests.rs`, `tests/create_message_tests.rs`, `tests/create_ai_tests.rs` all pin the current rollback behavior and would need re-verification against transaction semantics.

## What this issue is asking for

Wrap `stax create` (all its flows: name-only, `-m` commit-first, `--from`, `--below`, `--insert`) in a `Transaction` that covers the *entire* command, not just the `--insert` reparenting step — so:
- The whole operation becomes one undoable unit via `stax undo`, consistent with `sync`/`restack`/`split --file` (as of #838).
- `apply_insert_reparenting`'s metadata writes finally get before/after OID tracking in the op receipt, closing the last of #830's five sites for real diagnosability.
- The existing manual rollback closures (`rollback_after_commit`, `rollback_create`, `CreateBelowAutoStash`'s stash restoration) need to be reconciled with `Transaction`'s own drop/finish_err semantics rather than just left running alongside it — likely by having the manual rollback also call `tx.finish_err(...)` before returning, mirroring the pattern used in `split.rs`'s `try_or_rollback!` macro (see #838).

## Suggested approach

1. Read `create_branch_with_banner` and `run_commit_first` fully (both flows, all early-return/rollback sites) before touching anything — this is more archaeology than the other 4 sites combined.
2. Decide on an `OpKind` (new variant, or reuse an existing one like `OpKind::Split`/a new `OpKind::Create`).
3. Begin the transaction as early as is safe in each flow (probably right after validating placement flags, before any branch/commit mutation), and thread it through to every existing rollback call site, converting each to also call `tx.finish_err(...)`.
4. On the success path, plan + record the branch head and metadata ref for the created branch, plus (for `--insert`) each reparented child's metadata ref.
5. Full regression pass against the 5 existing create-focused test files listed above, plus new tests asserting the receipt is complete and that `stax undo` reverses a `--insert`/`--below`/commit-first create cleanly.

Related: #830, #834, #835, #838.

cesarferreira:
## Summary

Fixes #839.

#835/#838 tracked before/after metadata-ref OIDs at 4 of the 5 `parentBranchRevision` write sites fixed in #830/#834. The 5th, `apply_insert_reparenting` (`stax branch create --insert`), was deliberately deferred: `create.rs` had no `Transaction` anywhere, and scoping one to just the reparenting step would let `stax undo` reverse a child's metadata while leaving the newly created branch behind — worse than no receipt at all.

## What changed

Added a new `OpKind::Create` and threaded a `Transaction` through both of `create`'s independent flows:

- **`create_branch_with_banner`** (branch-first: plain name-only, or `--insert`/`--below` without `-m`) — plans + snapshots the new branch's head/metadata ref (and the reparented branch's metadata ref for `--below`) up front. Every existing manual rollback site already returns `Err`, so the two owning callers convert that into exactly one `finish_err`.
- **`run_commit_first`** (`-m` commit-first flow) — snapshots **lazily**, right before the first real mutation (branch ref creation). Earlier failures (pre-commit hook failure, non-zero commit exit, Ctrl+C, `rev-parse` failure) leave the transaction unsnapshotted and silently dropped — so pre-commit failures still produce no receipt, exactly matching current behavior. The ~6 existing rollback call sites are converted to a `rollback_and_finish!` macro that also calls `tx.finish_err(...)`.
- **`apply_insert_reparenting`** now plans/snapshots/records each reparented child's metadata ref per-child (`Transaction::snapshot` is incremental, so earlier children's before-OIDs are never clobbered) — closing the loop: its `parentBranchRevision` writes finally get before/after tracking in the op receipt.

**Deliberately untouched:** the plain `stax create <name>` fast path (`simple_explicit_empty_create`). Giving every branch creation a receipt would make `stax undo` nondeterministic in unrelated tests due to the op-ID timestamp's 1-second resolution, and that path has none of #830's write sites anyway.

## Test plan

- [x] 5 new tests in `tests/create_undo_tests.rs`: receipt completeness for `--insert`, `stax undo` after `--insert`, after `--below`, after commit-first `-m`, and a pre-commit-hook failure producing zero receipt.
- [x] Targeted regression suites (`create_*`, `split_tests`, `fold_tests`, `guard_rail_tests`, `sync_undo_tests`, `edge_cases_tests`, `ops::`/`receipt::`): 223 tests — pass
- [x] `make test` (full suite, 2620 tests) — pass
- [x] Independent verifier review confirmed all 6 risk areas hold structurally (correct lazy-snapshot boundary, fast path fully untouched, checked-out branch's head ref never planned so `undo` can't destroy user work, after-OIDs correctly resolved via subprocess not libgit2, `quiet: true` preserving existing stdout/stderr assertions, all 6 macro rollback-argument conversions verified byte-exact against the original code) — plus a manual scratch-repo exercise of a failing pre-commit hook, confirming zero receipt and a fully clean repo state on failure.

## Advisory finding (not introduced by this PR, flagging for follow-up)

Manual testing during verification found that **`stax undo` can target the wrong receipt when two creates land in the same second** — op IDs have only 1-second timestamp resolution (`src/ops/mod.rs:30`) with a lexicographic-string tiebreak on the random suffix (`src/ops/mod.rs:144`), so "latest by sort order" can disagree with "latest chronologically". Measured 4/6 disagreement in manual back-to-back trials. This is pre-existing infrastructure (the same hazard that's why the plain `bc <name>` fast path was deliberately excluded from receipt tracking above) — but this PR does make the commit-first `-m` path write a receipt on every invocation, without a matching exclusion, so it's newly exposed there. Will track as a separate follow-up rather than block this PR on it.

### DIFF
--- a/src/commands/branch/create.rs
+++ b/src/commands/branch/create.rs
@@ -28,6 +28,8 @@ use crate::commands::staging::{self, ContinueLabel, StagingAction};
 use crate::config::Config;
 use crate::engine::{BranchMetadata, Stack};
 use crate::git::GitRepo;
+use crate::ops::receipt::OpKind;
+use crate::ops::tx::Transaction;
 use crate::progress::LiveTimer;
 use crate::remote;
 use anyhow::{Context, Result, anyhow, bail};
@@ -666,7 +668,9 @@ pub fn run(
         return Ok(());
     }
 
-    create_branch_with_banner(
+    let mut tx = Transaction::begin(OpKind::Create, &repo, true)?;
+    let below_current = below_current_meta.as_ref().map(|_| current.as_str());
+    if let Err(e) = create_branch_with_banner(
         &repo,
         &config,
         &current,
@@ -675,7 +679,12 @@ pub fn run(
         insert,
         below_current_meta.as_ref(),
         !staging::is_staging_area_empty(workdir)?,
-    )?;
+        &mut tx,
+    ) {
+        record_create_after(&mut tx, workdir, &branch_name, below_current);
+        let _ = tx.finish_err(&e.to_string(), Some("create"), Some(&branch_name));
+        return Err(e);
+    }
     print_branch_name_warnings(&branch_name_result.warnings);
 
     // Stage/commit behavior:
@@ -683,12 +692,12 @@ pub fn run(
     // - StageMode::ExistingOnly (files already staged) => keep current index
     // - StageMode::None => no staging/committing
     if stage_mode != StageMode::None {
-        let workdir = repo.workdir()?;
-
         if (stage_mode == StageMode::All || needs_stage_all)
             && let Err(e) = staging::stage_all(workdir)
         {
             rollback_create_and_restore(&repo, &current, &branch_name, below_current_meta.as_ref());
+            record_create_after(&mut tx, workdir, &branch_name, below_current);
+            let _ = tx.finish_err(&e.to_string(), Some("stage"), Some(&branch_name));
             return Err(e);
         }
 
@@ -696,6 +705,10 @@ pub fn run(
             println!("{}", "Changes staged".dimmed());
         }
     }
+
+    record_create_after(&mut tx, workdir, &branch_name, below_current);
+    tx.set_head_branch_after(&branch_name);
+    tx.finish_ok()?;
 
     if config.ui.tips {
         println!(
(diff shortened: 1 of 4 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #840** 👈
    * **PR #842**
      * **PR #844**
        * **PR #845**
          * **PR #846**
            * **PR #847**

PR #842, opened 2026-08-30
fix(detach): track metadata refs so undo actually reverses detach

cesarferreira:
## Summary

First of a stack of fixes from a broader test-coverage audit prompted by the #830 series — after fixing 5 confirmed instances of unguarded `parentBranchRevision` writes and the diagnosability gaps around them, I ran `cargo llvm-cov` over the full suite and dug into the highest-risk, lowest-coverage areas of the codebase (git-mutating code, not just UI). This is the worst finding.

`stax detach`'s `Transaction` only planned/recorded each branch's HEAD ref via `plan_branch`/`record_after`. But `detach` never moves any branch's HEAD at all — its only mutation is rewriting `parentBranchName`/`parentBranchRevision` in each branch's *metadata* ref (`refs/branch-metadata/<branch>`). Since the tracked ref never changes, `oid_before == oid_after` for every entry the receipt records, so **`stax undo` after `stax detach` appeared to succeed but silently reversed nothing** — no error, a plausible-looking receipt, but the reparenting stayed in place.

## Fix

- Plan and record `refs/branch-metadata/<branch>` too, for the detached branch and every reparented child, via `plan_metadata_ref` + a subprocess-resolved `ref_oid` helper + `record_known_metadata_after` — metadata refs are written by a `git update-ref` subprocess, which libgit2's cached refdb can miss, so after-OIDs can't come from the same `record_metadata_ref_after`/libgit2 path used for before-OIDs. Same pattern used at every other site fixed this session (`sync.rs`, `split.rs`, `split_hunk/app.rs`, `branch/create.rs`).
- Routed the new `parentBranchRevision` through `GitRepo::resolve_child_parent_boundary` (merge-base tried first, then the tip, then the previously recorded value) instead of a raw `unwrap_or_else` fallback to an unverified branch tip — consistent with the #830 fixes. Behavior is unchanged in the normal case, since a merge-base is by definition an ancestor.

## Test plan

- [x] Three new tests, empirically confirmed to **fail against the unfixed code** (receipt missing the `@meta` entries; both undo tests showing `parentBranchName` still at its post-detach value even though `stax undo` exited 0) and pass after the fix: `detach_records_metadata_refs_in_the_op_receipt`, `detach_undo_restores_reparented_children_metadata`, `detach_undo_restores_the_detached_branchs_own_metadata`.
- [x] All 8 `detach_tests` (5 pre-existing + 3 new), plus `create_undo_tests`/`sync_undo_tests`/`split_tests`/`fold_tests` (41 tests) — pass
- [x] `make test` (full suite, 2620 tests) — pass
- [x] Independent verifier confirmed the `resolve_child_parent_boundary` candidate ordering/fallback is correct and behavior-preserving in the happy path, and that subprocess-resolved recording (not libgit2) is used throughout.

### DIFF
--- a/src/commands/detach.rs
+++ b/src/commands/detach.rs
@@ -4,6 +4,8 @@ use crate::ops::receipt::OpKind;
 use crate::ops::tx::Transaction;
 use anyhow::Result;
 use colored::Colorize;
+use std::path::Path;
+use std::process::Command;
 
 pub fn run(branch: Option<String>, yes: bool) -> Result<()> {
     let repo = GitRepo::open()?;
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #840**
    * **PR #842** 👈
      * **PR #844**
        * **PR #845**
          * **PR #846**
            * **PR #847**

PR #844, opened 2026-08-30
fix(edit): track metadata ref for undo, and fix a real cp bug in the rebase editor

cesarferreira:
## Summary

Second in the stack of fixes from the broader test-coverage audit (see #842 for the first).

`stax edit`'s `Transaction` only tracked the branch's HEAD ref, never its metadata ref, even though a successful edit rewrites `parentBranchRevision`. `stax undo` after `stax edit` correctly restored the branch's commits (HEAD was tracked) but left `parentBranchRevision` at its post-edit value instead of the pre-edit one — same family of bug as `detach.rs` (#842).

## A second, unrelated bug found and fixed along the way

`stax edit`'s per-commit action selection requires a real terminal — even `--yes` doesn't help, confirmed by an existing test asserting exactly that. So fixing the metadata-tracking gap required extracting the non-interactive tail of `run()` into a separate, directly-testable `apply_edit_plan()` function first.

While doing that extraction, I found a second, completely unrelated, pre-existing bug: the `GIT_SEQUENCE_EDITOR` command string was `cp '<todo>' "$1"`. Git's editor launcher always appends the todo-file path as `"$@"` to whatever command is configured — so the explicit `"$1"` made `cp` receive the destination path twice (`cp src dest dest`), which fails with `cp: target ... Not a directory`. **This means `stax edit`'s actual rebase step never completed successfully for any real invocation.** I independently reproduced this in isolation outside the test suite to confirm it, and traced why it survived: the one existing test that performs a real rebase drives `git rebase -i` directly instead of going through `stax edit`'s own binary, and happens to use the correct form by coincidence — so this exact code path in the shipped binary was never exercised by any test.

## Fix

- `apply_edit_plan()` now plans/records `refs/branch-metadata/<branch>` via the same subprocess-resolved `ref_oid` + `record_known_metadata_after` pattern used at every other site this session (metadata refs are written via `git update-ref`, which libgit2's cached refdb can miss).
- Fixed the `cp` command string to `cp '<todo>'` (no `"$1"`).
- Added a real end-to-end test driven through a pty (not just unit tests against the extracted function, which bypass the terminal) that exercises the actual interactive `stax edit` binary — this is the test that would have caught both bugs, and does catch the `cp` regression when reverted.

## Test plan

- [x] Two new unit tests against `apply_edit_plan` directly, empirically confirmed to fail against the unfixed metadata-tracking code and pass after.
- [x] One new real end-to-end pty-driven test (`edit_drop_via_real_interactive_session_updates_metadata_and_supports_undo`), empirically confirmed to fail with the exact real-world `cp` error when the `cp` fix is reverted, and pass with both fixes applied — including a full `stax undo` round-trip verified at the CLI level.
- [x] All `edit_tests`/`commands::edit::tests` (9 tests) plus `detach_tests`/`sync_undo_tests` (26 tests) — pass
- [x] `make test` (full suite, 2626 tests) — pass
- [x] Independent verifier review: confirmed the `cp` bug three separate ways (raw git semantics, reverting just that line against the new tests, and a manual pty-driven CLI run), confirmed the extraction is behavior-preserving for the interactive path, and confirmed the borrow-trap fix (`meta.parent_branch_name.clone()`) doesn't change behavior elsewhere in `run()`.

### DIFF
--- a/src/commands/edit.rs
+++ b/src/commands/edit.rs
@@ -8,6 +8,7 @@ use console::Term;
 use dialoguer::Select;
 use dialoguer::theme::ColorfulTheme;
 use std::io::Write;
+use std::path::Path;
 use std::process::Command;
 
 /// Commit info parsed from `git log`.
@@ -70,7 +71,7 @@ pub fn run(yes: bool, no_verify: bool) -> Result<()> {
     // Find the parent branch boundary
     let meta = BranchMetadata::read(repo.inner(), &current)?
         .ok_or_else(|| anyhow::anyhow!("Branch '{}' is not tracked by stax", current))?;
-    let parent = &meta.parent_branch_name;
+    let parent = meta.parent_branch_name.clone();
     let child_branches = stack.descendants(&current);
 
     // Get commits between parent and HEAD (oldest first)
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #840**
    * **PR #842**
      * **PR #844** 👈
        * **PR #845**
          * **PR #846**
            * **PR #847**

PR #845, opened 2026-08-30
fix(restack): verify ancestry when recovering metadata after a manual rebase --continue

cesarferreira:
## Summary

Third in the stack of fixes from the broader test-coverage audit (see #842, #844 for the first two).

When `stax restack --continue` recovers from a rebase that was completed *manually* (`git rebase --continue` instead of `stax continue`), it blindly stamped the parent branch's current tip as `parentBranchRevision` — same #830 pattern fixed at five other sites earlier this session. If the parent advanced again during the pause window between the conflict and the manual completion, the child was never actually rebased onto that newer tip — poisoning the boundary for the next restack.

## Fix

Route the value through the existing `GitRepo::resolve_child_parent_boundary` helper (added in the original #830 fix): only trust the parent's current tip if it's verified to be an ancestor of the child, else keep the previously recorded (still-valid) boundary. Behavior is unchanged in the normal (non-race) case, since the resolved value equals the parent's tip whenever it really is an ancestor — confirmed via case analysis and by an independent verifier.

## Test plan

- [x] New test reproduces the exact race (conflict → resolve → manual `git rebase --continue`, bypassing stax's own metadata write entirely → parent advances *again* → `stax restack --continue`), empirically confirmed to fail against the unfixed code (`Recorded boundary ... is not an ancestor` panic) and pass after.
- [x] `continue_tests::` (13 tests) — pass
- [x] `make test` (full suite, 2627 tests) — pass, confirmed independently twice
- [x] Independent verifier confirmed structurally: the candidate/fallback ordering is correct for this recovery context (no second "old parent" candidate exists here, unlike the reparent-on-deletion site), and the removed guard is provably redundant with the new one via full case analysis over ancestry × equality.

### DIFF
--- a/src/commands/restack.rs
+++ b/src/commands/restack.rs
@@ -56,13 +56,24 @@ pub fn run(
             {
                 if let Some(meta) = BranchMetadata::read(repo.inner(), failed_branch)?
                     && let Ok(actual_parent_rev) = repo.branch_commit(&meta.parent_branch_name)
-                    && meta.parent_branch_revision != actual_parent_rev
                 {
-                    let updated = BranchMetadata {
-                        parent_branch_revision: actual_parent_rev,
-                        ..meta
-                    };
-                    updated.write(repo.inner(), failed_branch)?;
+                    // `actual_parent_rev` is only a valid boundary if `failed_branch` was
+                    // actually rebased onto it. If the parent advanced again during the
+                    // pause window (between the conflict and a manual `git rebase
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #840**
    * **PR #842**
      * **PR #844**
        * **PR #845** 👈
          * **PR #846**
            * **PR #847**

PR #846, opened 2026-08-30
fix(split): verify ancestry and track metadata refs in the interactive split TUI

cesarferreira:
## Summary

Fourth in the stack of fixes from the broader test-coverage audit (see #842, #844, #845 for the first three).

`src/tui/split/app.rs` (the old, non-hunk interactive `stax split`) had two real bugs, both already fixed in sibling files this session:

1. **Unguarded ancestry write** (same #830 pattern): every new split branch's `parentBranchRevision` was stamped with the live current tip of its parent branch, with no verification that the new branch's commits (captured earlier, at `SplitApp::new()` time) actually descend from that tip. If the parent branch advances between opening the interactive TUI and confirming the split, this poisons the boundary.
2. **Zero op-receipt tracking at all**: `tx.plan_branches()` only planned branch HEAD refs, and there was no `record_after` call anywhere before `finish_ok()` — not even branch HEAD after-OIDs were ever recorded. `stax undo` after a plain (non-hunk) `stax split` was likely completely non-functional, not just partially, unlike the other sites fixed this session.

## Fix

Route each write through `merge_base` + the existing `GitRepo::resolve_child_parent_boundary` helper (verify ancestry, fall back to a safe value otherwise), plan metadata refs for every affected branch, and record after-OIDs (both branch head and metadata ref, subprocess-resolved) for every branch touched.

Behavior is unchanged in the normal case: a merge-base always equals the parent's tip whenever that tip really is an ancestor, which is true for every split point after the first and for the current-branch reparent, and true for the first split point whenever nothing advanced mid-session — confirmed both by reasoning and by an independent verifier.

## Test plan

- [x] Two new tests directly in `src/tui/split/app.rs`'s test module, empirically confirmed to fail against the unfixed code (one on a boundary failing an `is_ancestor` check, one on missing `@meta` receipt entries) and pass after.
- [x] `tui::split::app::tests::` + `split_tests::`/`split_hunk_tests::` (34 tests) — pass
- [x] `make test` (full suite, 2629 tests) — pass
- [x] Independent verifier review: confirmed candidate ordering is correct at both write sites, confirmed every planned ref has a matching after-OID recording (no `after_recorded == false` gaps), confirmed this file is genuinely independent from the already-fixed `tui/split_hunk/app.rs`, and independently re-confirmed both bugs via a negative control (reverting the fix and re-observing the exact failures).

### DIFF
--- a/src/tui/split/app.rs
+++ b/src/tui/split/app.rs
@@ -4,6 +4,8 @@ use crate::git::GitRepo;
 use crate::ops::receipt::{OpKind, PlanSummary};
 use crate::ops::tx::{self, Transaction};
 use anyhow::{Context, Result};
+use std::path::Path;
+use std::process::Command;
 
 /// A commit to display in the split UI
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #840**
    * **PR #842**
      * **PR #844**
        * **PR #845**
          * **PR #846** 👈
            * **PR #847**

PR #847, opened 2026-08-30
test(reorder): pin the metadata-ref after-OID invariant, confirmed safe not buggy

cesarferreira:
## Summary

Fifth and final in the stack of fixes from the broader test-coverage audit (see #842, #844, #845, #846 for the first four).

## What was investigated

`application/reorder.rs` and `application/move_subtree.rs` both call `Transaction::record_metadata_ref_after`, which resolves a metadata ref's after-OID via libgit2 (`repo.find_reference`) even though metadata refs are written by a `git update-ref` **subprocess** — the exact pattern that turned out to be a real bug at 6 other sites fixed this session. Flagged during the audit as a "strong candidate, not confirmed via repro."

## What I found

**Not a bug.** Empirically confirmed via a real reorder and move-subtree operation, in both normal (loose refs) and adversarially-forced (`git pack-refs --all` beforehand) modes, that the recorded after-OIDs always match a fresh subprocess `git rev-parse`.

Root cause of why this differs from the 6 confirmed-buggy sites: `git update-ref` always leaves a loose ref file on disk, and libgit2 reads loose refs straight off disk, ahead of the stat-cached packed-refs backend — so there's no staleness window for a pure write. The staleness risk that bit the other 6 sites is specific to **deletion** (`git update-ref -d`, which does rewrite packed-refs and can leave libgit2 serving a stale cached value) — something neither `reorder.rs` nor `move_subtree.rs` ever does to a metadata ref.

I also checked the unguarded-ancestry-write pattern (bug class 1 from #830) at both files' `parent_branch_revision` writes — also not applicable here, since both writes happen immediately after a real, successful rebase onto that exact parent, with no metadata-only-reparent gap.

## What this PR does

No code fix — because none is needed. Instead:
- A regression test (`reorder_and_move_record_metadata_after_oids_that_match_git`, run in both loose and packed-refs modes) pins this invariant, so a future change to the write path here (e.g. switching to a delete-then-recreate strategy) would fail loudly instead of silently reintroducing the bug.
- A doc comment on `Transaction::record_metadata_ref_after` explaining precisely when it's safe (write-only callers) versus when it's not (callers that delete the ref, which must use the subprocess-resolved `record_known_metadata_after` pattern instead) — so this doesn't need re-investigating a third time.

## Test plan

- [x] New test passes in both loose-refs and packed-refs modes, confirming the invariant holds.
- [x] `application_operation_tests::` + `ops::` (94 tests) — pass
- [x] `make test` (full suite, 2630 tests) — pass

### DIFF
--- a/src/ops/tx.rs
+++ b/src/ops/tx.rs
@@ -231,6 +231,17 @@ impl Transaction {
     /// may be absent (e.g., metadata was deleted) — that's recorded as
     /// `oid_after = None`, which `stax undo` handles by re-creating the ref
     /// from `oid_before`.
+    ///
+    /// Safe for callers that only ever *write* a branch's metadata ref (via
+    /// `BranchMetadata::write` -> `git::refs::write_metadata`, a `git update-ref`
+    /// subprocess): `update-ref` always leaves a loose ref file, and libgit2 reads
+    /// loose refs straight off disk, ahead of the stat-cached packed-refs backend,
+    /// so there's no staleness window to worry about here (verified empirically,
+    /// including with the ref pre-packed via `git pack-refs`). Callers that instead
+    /// *delete* a metadata ref (`git update-ref -d`, which does rewrite packed-refs)
+    /// must resolve the after-OID via a subprocess `git rev-parse --verify <ref>`
+    /// helper and `record_known_metadata_after` instead — see `sync.rs`, `split.rs`,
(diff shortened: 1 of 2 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #840**
    * **PR #842**
      * **PR #844**
        * **PR #845**
          * **PR #846**
            * **PR #847** 👈

PR #852, opened 2026-08-31
fix(github): route GitHub API calls through the configured HTTP proxy

geoHeil:
## The problem

On a network where egress requires an HTTP `CONNECT` proxy, every GitHub API call fails at connect time:

```
$ st ss
Submitting stack...
  ✓ Fetching from origin...             done 2.977s
Error: Failed to list PRs by head: Service Error: client error (Connect): client error (Connect): deadline has elapsed
```

The fetch succeeds because git honours `http_proxy`. The API call immediately after it does not, because octocrab has no proxy support: `build()` hardcodes `HttpsConnectorBuilder` over hyper-util's `HttpConnector`, which never consults proxy configuration, and `grep -ri proxy` over the crate returns nothing. No env var or `config.toml` key can fix that from outside — a client with no proxy code cannot use a proxy.

The GitLab and Gitea paths were unaffected: they use `reqwest`, which does read `HTTPS_PROXY`/`NO_PROXY`.

## The fix

When `ALL_PROXY`/`HTTPS_PROXY`/`HTTP_PROXY` (either case) is set, `GitHubClient::new_with_auth` builds the `Octocrab` on a reqwest-backed `tower` service — the same client stack the other forges already use. reqwest handles `CONNECT` tunnelling, `NO_PROXY` exclusions, proxy credentials, and platform certificate verification, which also covers proxies that terminate TLS with a CA from the OS trust store.

**With no proxy variable set, nothing changes** — the existing octocrab builder path is taken verbatim, so unproxied users are unaffected.

`with_service()` moves the builder into the `NoConfig` type state, whose `build()` applies nothing but `MapResponseBodyLayer`, so `src/github/transport.rs` reapplies octocrab's own layers in octocrab's order:

| Layer | Why it can't be skipped |
|---|---|
| `ExtraHeadersLayer` | GitHub rejects requests without a `User-Agent` |
| `BaseUriLayer` | handlers build relative URIs |
| `AuthHeaderLayer` | `Authorization`, plus the authority check that keeps the token from following a redirect off-host |

Retry is a loop over the buffered request body rather than `RetryConfig`, whose `Policy` impl is bound to `hyper_util::client::legacy::Error` — an error type a non-hyper service cannot construct. Filed upstream as [XAMPPRocky/octocrab#936](https://github.com/XAMPPRocky/octocrab/issues/936), asking for connector injection so this module can shrink to a few lines.

## Also in here

- **Better connect diagnostics.** A connect failure previously fell through to the auth branch of `enrich_api_error` and suggested the token was expired. It now reports whether a proxy was in play and which variable named it, with `user:password@` redacted.
- **A latent panic fixed.** reqwest is built with `rustls-no-provider`, which panics unless a crypto provider is installed. `cli::run` installs one, so the CLI was fine, but a client built outside that entry point (tests, library use) panicked. The transport now installs one defensively.
- **Deterministic tests.** `scripts/native-tests.sh` and the Makefile's native runs now scrub proxy variables alongside the tokens, because a proxy inherited from the developer's shell reroutes the loopback mock servers the tests talk to. `scripts/native-tests-tests.sh` asserts the scrub.
- **Docs.** New "HTTP proxies" section in `docs/configuration/index.md`.

## Tests

Six new tests in `src/github/transport.rs` and `src/github/client.rs`:

- a wiremock round trip through the custom transport, asserting `User-Agent` and `Authorization` arrive and the relative URI resolves against the base — i.e. the hand-built layer stack behaves like octocrab's;
- retry recovers from a single 500 (2 requests), and surfaces the error with retries disabled (1 request);
- proxy-variable resolution order and blank-value handling;
- credential redaction in the hint;
- connect errors get the proxy hint and not the auth hint.

Verified end-to-end against real GitHub through an HTTP `CONNECT` proxy: REST reads and a GraphQL POST both succeed where they previously died at connect.

`cargo check --all-targets`, `./scripts/lint.sh full` and `cargo fmt --check` are clean. The full suite was run natively with a proxy deliberately left set, so the new transport was the one under test: **2649/2655 pass**. The 6 failures are pre-existing and unrelated — they fail identically on `main` at this commit: 4 PTY-driven tests need Linux `script` semantics (which is why `make test` prefers Docker on macOS), and 2 read a global `remote.base_url` from the developer's own stax config. CI, which runs the suite in the Linux image with a clean environment, should be green.

## Follow-up, deliberately not in this PR

Collapsing GitHub onto the same reqwest stack as the other forges would delete the dual transport and the small divergences it carries (user-agent `stax` vs `octocrab`; no write timeout on the proxy path). Bigger blast radius, so it belongs in its own change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3593,6 +3593,8 @@ dependencies = [
  "fuzzy-matcher",
  "getrandom 0.4.3",
  "git2",
+ "http",
+ "http-body-util",
  "indicatif",
  "maud",
  "octocrab",
@@ -3607,6 +3609,7 @@ dependencies = [
  "termimad",
  "tokio",
  "toml",
+ "tower",
  "update-informer",
  "wiremock",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,11 @@ git2 = { version = "0.21", default-features = false, features = ["https"] }
 
 # GitHub API
 octocrab = "0.54.1"
+# Proxy-aware GitHub transport (octocrab's own hyper client ignores HTTPS_PROXY;
+# see src/github/transport.rs). Versions match what octocrab/reqwest already pull in.
+http = "1"
+http-body-util = "0.1"
+tower = { version = "0.5", default-features = false, features = ["util"] }
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "net", "sync", "signal", "time"] }
 futures-util = { version = "0.3", default-features = false, features = ["std"] }
 
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ test-local-fast:
 		./scripts/native-tests.sh; \
 	else \
 		threads="$${NEXTEST_TEST_THREADS:-num-cpus}"; \
-		env -u GITHUB_TOKEN -u STAX_GITHUB_TOKEN -u GH_TOKEN STAX_DISABLE_UPDATE_CHECK=1 STAX_TEST_TMPDIR="$$(pwd)/.test-tmp" TMPDIR="$$(pwd)/.test-tmp" NEXTEST_TEST_THREADS="$$threads" RUST_MIN_STACK=4194304 cargo nextest run --cargo-profile "$(NATIVE_CARGO_PROFILE)"; \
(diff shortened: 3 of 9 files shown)


geoHeil:
Follow-up raised: **#853** makes this transport the only one, deleting the proxied/unproxied branch that this PR deliberately kept.

**Merge order: this PR (#852) first, then #853.**

Because I have no push access here, #853's base branch could not live in this repo, so GitHub cannot render it as a true stacked PR — its diff currently includes this PR's two commits as well. Once #852 lands I will rebase #853 onto `main` and force-push, after which it shows only its own 51-line refactor. The isolated diff is viewable now at geoHeil/stax#2.

#853 is reviewable independently of that mechanics if you want to weigh the two behaviour changes it calls out (`User-Agent` becomes `stax` for unproxied users, and a 30s write timeout becomes a 60s total timeout, since reqwest has no write-timeout equivalent).

cesarferreira:
merged this one, when #853 is ready I'll merge it

PR #855, opened 2026-09-01
feat(merge): add --ignore-failed-ci to override a red CI rollup

### LINKED ISSUE #851: feat: merge blocks on optional failed CI checks with no override
## Summary

Stax refuses to merge when any CI check has failed, even when GitHub considers the failed check optional and the pull request mergeable. There is no documented merge flag to distinguish required checks from optional checks or explicitly bypass Stax readiness. I think it would be nice with an option to bypass the block on CI failures and still merge.

## Environment
- Stax 0.109.0
- GitHub repository with native stacked PRs enabled
- macOS

## Reproduction
1. Open a PR whose required checks pass.
2. Let an optional integration check fail. In this case, Vercel reported Deployment was blocked
3. Confirm that GitHub reports the pull request as mergeable, with mergeStateStatus UNSTABLE rather than BLOCKED.
4. Run:

  `st merge --method merge --no-delete --no-sync`

Stax exits with:
```
CI failed
Fix the issue and run stax merge again.
```

The available flags do not provide an escape hatch:

- `--no-wait` only changes handling of pending checks.
- `--yes` only skips confirmation.
- `--remote` changes execution location.

## Expected behavior

Ideally, Stax should distinguish GitHub-required checks from optional failing checks and allow a merge when GitHub branch protection permits it.

Alternatively, provide an explicit, clearly named override such as `--ignore-failed-ci` or `--force` for cases where the caller intentionally wants GitHub to make the final branch-protection decision.

cesarferreira:
## Motivation

Closes #851.

`stax merge` refused to merge whenever the forge reported the CI rollup as failed, even when the failing check was optional and GitHub itself considered the PR mergeable (`mergeStateStatus: UNSTABLE`, not `BLOCKED`). The reporter hit this with a Vercel preview deployment that reported "Deployment was blocked" while every required check passed. There was no escape hatch: `--no-wait` only changes pending-check handling, `--yes` only skips the confirmation prompt, and `--remote` only changes where the merge executes. The run just died with:

```
CI failed
Fix the issue and run stax merge again.
```

## What changed

Adds `stax merge --ignore-failed-ci`, an explicit, narrowly-scoped override.

The readiness gate lived in `PrMergeStatus::is_ready()` / `is_blocked()` / `status_text()`, which conflated "CI is red" with "this PR must not merge". Rather than duplicate that logic per merge mode or mutate status objects in place, this introduces a `ReadinessPolicy { ignore_failed_ci: bool }` value that those three methods now take, and threads it from the CLI down through the merge family:

- `src/forge/model.rs` — new `ReadinessPolicy` (`Copy`, `Default` = waive nothing) with `strict()` / `ignoring_failed_ci()`; the three gate methods take it.
- `src/commands/merge_shared.rs` — `blocked_reason` and `wait_for_pr_ready` take the policy; new `warn_ignored_ci_failure` helper.
- `src/commands/merge_stack.rs` — `tip_blocker` and `wait_for_tip_ready` take the policy. `downstack_blocker` is deliberately unchanged: it already ignores CI by design, because lower PRs in a `--stack` range land indirectly through the tip merge.
- `src/commands/{merge,merge_stack,merge_when_ready,merge_remote}.rs` — `run()` accepts and forwards the policy.

The flag applies to the default cascade, `--stack`, `--remote`, and `--when-ready` — every mode where stax gates locally.

### Semantics, deliberately narrow

`--ignore-failed-ci` waives exactly one thing: a **failed** CI rollup. It still blocks on draft PRs, requested changes, merge conflicts, and closed PRs. A **pending** rollup is still waited on rather than waived — that is `--no-wait`'s job, and conflating the two would make the flag a blunt "merge anything" instrument. If a run turns red while stax is waiting, the next poll now returns `Ready` instead of `Failed`.

Making `status_text()` policy-aware is load-bearing, not cosmetic. It checks CI failure *before* the conflict branch, so without the policy a conflicting PR merged with `--ignore-failed-ci` would have reported "CI failed" — telling the user to fix the exact thing they just overrode. There is a unit test and an integration test pinning that it says `Has conflicts` instead.

The override is never silent: when it actually skips a red rollup, stax prints a warning naming the flag and noting that the forge's branch protection is still the final gate. If GitHub refuses the merge, the existing `merge_pr` error path surfaces that unchanged. Conversely, when CI failure *does* block a merge, the failure summary now points at the flag, so the escape hatch is discoverable at the moment it is needed.

`--queue` rejects the flag at parse time via `conflicts_with_all`. `merge_queue.rs` has no local readiness gate at all — the forge's merge queue decides — so a silently-ignored flag would be worse than a clap error.

### Why not required-vs-optional check detection

Issue #851 asked for either (a) distinguishing GitHub-required from optional checks, or (b) an explicit override. This implements (b).

Option (a) is a materially larger change and is not attempted here. stax never fetches `mergeStateStatus` — `src/github/pr.rs` maps only `mergeable` (`MERGEABLE` / `CONFLICTING` / `UNKNOWN`) via `graphql_mergeable_state`. Reading `UNSTABLE` vs `BLOCKED` needs a GitHub preview `Accept` header and has no GitLab or Gitea equivalent, so it would fork readiness semantics per forge. `--ignore-failed-ci` is the forge-agnostic fix, and it composes with (a) later: required-check awareness would shrink how often the flag is needed without changing what it means.

## Verification

Evidence is from the **full local gate** on Linux, not the scoped draft gate — the change touches `src/forge/model.rs`, shared core readiness logic consumed by the whole merge family, which triggers the high-risk exception in `AGENTS.md`.

- `cargo check`, `cargo fmt --check`, `git diff --check` — clean.
- `make lint` (full, `--all-targets --all-features`) — clean.
- `make test` — **2674 passed, 0 failed, 0 skipped**.

New tests:

- **Unit** (`src/github/pr.rs`): override makes a CI-failed-but-clean PR ready; does not waive draft / changes-requested / conflicts / closed; pending CI is neither ready nor blocked under the override.
- **Unit** (`src/commands/merge_stack.rs`): `tip_blocker` ignores CI failure under the override, still blocks pending CI and drafts.
- **Unit** (`src/commands/merge_shared.rs`): `blocked_reason` reports `Has conflicts`, not `CI failed`, under the override.
- **Integration** (`tests/merge_ignore_failed_ci_tests.rs`, 5 end-to-end tests via the real binary against a wiremock forge): each asserts on whether `PUT /pulls/1/merge` was actually called, so a passing assertion means the merge really did or did not happen. Covers the happy path, the no-flag regression (still blocks, and surfaces the hint), and the three bad paths (draft, changes-requested, conflicts).
- **Clap surface**: flag parses; `--queue --ignore-failed-ci` is rejected.

## Risk

Low, and the risk is concentrated in the signature change rather than the behavior change. Default behavior is unchanged by construction: `ReadinessPolicy::default()` waives nothing, and every existing caller passes `strict()`, so a red CI rollup blocks exactly as before unless the user opts in.

The three `PrMergeStatus` methods gained a parameter. They had no callers outside the merge family and their own tests (`src/commands/ready.rs` and the TUI read `PrMergeStatus` fields directly), so this is contained — and the compiler enforces that every future call site makes a policy decision, rather than leaving a policy-free overload that silently reintroduces this bug.

The genuine residual risk is behavioral, not mechanical: a user can now merge a PR with failing required checks if branch protection does not enforce them. That is the requested capability, it requires an explicit flag, it warns on use, and the forge remains the final authority.

## Docs

Updated `README.md`, `docs/workflows/merge-and-cascade.md` (including what the flag does *not* waive), `docs/commands/reference.md`, and `skills.md`. `CHANGELOG.md` is untouched — it is generated by git-cliff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,7 @@ st merge --stack          # GitHub/GitLab: preserve lower PR/MR merged state thr
 st merge --stack --full   # stack-merge the full stack even from the middle
 st merge --remote         # merge remotely on GitHub while you keep working
 st merge --all            # merge the whole stack regardless of position
+st merge --ignore-failed-ci  # merge despite a red CI rollup (e.g. only optional checks failed)
 ```
 
 On GitHub, if the current stack is a confirmed-enabled native GitHub Stack and
@@ -366,7 +367,7 @@ These rules apply to AI-generated PR titles and bodies from `generate` and `subm
 | `st upstack submit` | Submit current branch and descendants, chaining temporary publish heads when needed |
 | `st reviews --stack [--json]` | Stack-wide review/comment inbox, including inline file/line locations on GitHub (`st comments` remains the current-PR view) |
 | `st next` | Move to the next unmerged branch upstack; fork choices are deterministic |
-| `st merge` | Cascade-merge from bottom to current (`--when-ready`, `--downstack-only`/`--ds`, `--stack`, `--stack --full`, `--remote`, `--all`) |
+| `st merge` | Cascade-merge from bottom to current (`--when-ready`, `--downstack-only`/`--ds`, `--stack`, `--stack --full`, `--remote`, `--all`, `--ignore-failed-ci`) |
 | `st ready` | Interactive PR readiness TUI — CI, review approval, and merge state for unmerged tracked PRs; auto-refreshes every 15s and drops remotely merged PRs without cleaning up their local branches (`--current`/`--stack` for current stack, `--plain` for a static table, `--json` for machine-readable readiness schema) |
 | `st board` / `st home` | Interactive repository dashboard (GitHub only) — PULL REQUESTS / ISSUES tabs with a detail pane, inline diff/comments, label editing, draft toggle, and API-only merge (`--limit`, `--tab`, `--interval`, `--plain` for static tables) |
 | `st ci` / `st ci --oneline` | Live CI status for each PR head — full per-check table, or one compact line per branch across the stack; GitHub statuses and check runs are fetched across all pages before roll-up |
--- a/docs/commands/reference.md
+++ b/docs/commands/reference.md
@@ -334,6 +334,7 @@ st completions elvish
 - `--timeout <minutes>` — positive whole minutes (default: 30); zero is rejected
 - `--interval <seconds>` — positive whole seconds (default: 15) for `--when-ready`, `--remote`, `--queue`, and `--stack --when-ready`; zero is rejected
 - `--no-wait` / `--no-sync` / `--no-delete` / `--quiet`
+- `--ignore-failed-ci` — merge even when CI checks have failed; still blocks on draft, changes-requested, conflicts, and closed. Incompatible with `--queue`. Use when only optional checks failed and branch protection does not require them.
 
 For `--queue`, the timeout is a real deadline: stax caps the final sleep to the remaining budget and does not start another forge status poll once the deadline is reached.
 
--- a/docs/workflows/merge-and-cascade.md
+++ b/docs/workflows/merge-and-cascade.md
@@ -6,7 +6,7 @@ How to merge an entire stack safely.
 
 Cascade-merges PRs from the bottom of your stack up to your current branch. For each PR, stax:
 
-1. Waits for readiness (CI + approvals + mergeability) unless `--no-wait`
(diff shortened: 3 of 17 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #855** 👈

PR #868, opened 2026-09-07
fix(skills): repair broken symlinks and keep per-harness failures non-fatal

cesarferreira:
## Motivation

`stax skills update` aborted the entire run with a confusing error when a harness skill directory was a symlink that no longer resolved:

```
Error: Failed to create directory /Users/me/.codex/skills/stax: File exists (os error 17)
```

Two separate problems produced that outcome:

1. **A dangling symlink is unrecoverable.** Before writing, we call `std::fs::create_dir_all` on the skill directory. `mkdir` returns `EEXIST` because the symlink occupies the name, and `create_dir_all` then decides whether to swallow that by testing `path.is_dir()` — which *follows* the link. For a dangling link that is `false`, so the `EEXIST` surfaces as a hard error. (Had the target existed, the write would have gone through the symlink fine.) This is easy to hit with a dotfiles symlink farm whose target directory isn't present on the current machine.

2. **One bad harness killed the whole run.** The `?` inside the per-harness loop propagated out of `run_update_with`, so the four harnesses after the failing one were silently never updated.

- Detect broken symlinks at both the skill-directory and `SKILL.md` positions before writing. Detection uses `is_symlink()` (does not follow) combined with `is_dir()`/`is_file()` (do follow), so "is a link but does not resolve to a usable target" is exactly the case we repair.
- Repair removes only the link via `remove_file`, never the target, then creates a real path. A visible yellow line reports the replacement so a link disappearing from the user's dotfiles isn't silent.
- **Symlinks that resolve correctly are left completely untouched and written through**, preserving intentional symlink farms. This is locked in by a dedicated test rather than left to reasoning.
- `--dry-run` stays strictly read-only: it reports `would replace broken symlink …` and performs no `remove_file`/`create_dir_all`/`write`.
- Per-harness failures are now non-fatal. Each prints `✗ <Harness> failed: <full anyhow chain>`, the loop continues, the summary reports the failure count, and the command exits non-zero via a final `bail!`.
- `maybe_install_skills` in `shell_setup.rs` downgrades the new error to a yellow warning, matching the existing precedent two lines above, so one bad harness path can't fail the whole `stax setup` wizard.
- Directory-symlink ancestor scanning is bounded by `$HOME`, so nothing at or above the home directory is ever a repair candidate.
- Non-symlink obstacles (e.g. a regular file sitting where the skill directory should be) are reported as failures and never deleted.

Docs: added one sentence to `docs/commands/reference.md` covering both behaviours. `README.md` and `skills.md` need no change — no command, flag, or workflow surface changed.

## Verification

Scoped draft gate, run locally on this branch head:

- `cargo check` — pass
- `make lint` (full, not just `lint-fast`) and `cargo fmt --check` — pass
- `git diff --check` — pass
- `cargo nextest run --lib commands::skills::tests::` — 27 passed, including 4 new unit tests for `detect_broken_links` / `broken_dir_link`
- `cargo nextest run skills_symlink_tests:: skills_selection_tests::` — 11 passed

New integration tests in `tests/skills_symlink_tests.rs` (registered in `tests/all_tests.rs`, `#[cfg(unix)]`) cover the happy path, the bad path, and the edge cases:

| Test | Covers |
|---|---|
| `skills_update_replaces_dangling_skill_dir_symlink` | The reported bug; asserts the output never contains `File exists` |
| `skills_update_writes_through_valid_dir_symlink` | A working symlink stays a symlink and is written through |
| `skills_update_continues_after_harness_failure_and_exits_nonzero` | A regular-file obstacle on the first harness still lets the other four update; exit is non-zero and the obstacle is not deleted |
| `skills_update_dry_run_leaves_broken_symlink_intact` | Dry run mutates nothing |
| `skills_update_replaces_broken_skill_file_symlink` | Broken `SKILL.md` link whose target parent is missing (a plain `fs::write` would fail `ENOENT`) |

Also verified against a real affected machine, which had two dead links (`~/.codex/skills/stax` and `~/.claude/skills/stax`, both pointing into a directory that no longer exists). `--dry-run` reported both; the real run replaced them and updated all five harnesses.

`cargo clippy --all-targets` reports two pre-existing `items after a test module` errors in `src/web/routes.rs` and `src/web/templates.rs`, unrelated to this change and present on `main` — which is why the repo's own `make lint` scopes clippy to lib and bins.

## Risk

Low and contained to `stax skills update` / `stax setup`. The one destructive operation is `remove_file` on a path already proven to be a symlink that does not resolve, so no reachable data can be lost. The new non-zero exit on partial failure is a deliberate behaviour change: `cli/mod.rs` maps it to exit code 1, and the two other callers (`doctor.rs` repair actions, `propose_skills_update`) propagate it consistently with every other failure in those paths.

### DIFF
--- a/docs/commands/reference.md
+++ b/docs/commands/reference.md
@@ -203,7 +203,7 @@ when it points at the same commit. `--all` conflicts with `--parent` and
 | `st open` | Open repository in browser |
 | `st demo` | Interactive tutorial — no auth or repo required |
 
-`st skills update` fetches the remote body and compares each fully rendered harness file byte-for-byte, rewriting instructions that differ even when the installed package-version marker matches. It leaves byte-identical files untouched; `--dry-run` reports the same decision without writing. `st skills list` is intentionally local-only: its current/stale indicator compares the installed package-version marker with this stax binary and does not verify fetched content.
+`st skills update` fetches the remote body and compares each fully rendered harness file byte-for-byte, rewriting instructions that differ even when the installed package-version marker matches. It leaves byte-identical files untouched; `--dry-run` reports the same decision without writing. `st skills list` is intentionally local-only: its current/stale indicator compares the installed package-version marker with this stax binary and does not verify fetched content. When a symlink at the skill directory or file position no longer resolves, `st skills update` removes it and replaces it with a real path (working symlinks are left intact and written through); if a harness cannot be written for any other reason it is reported and skipped so the remaining harnesses still update, and the command exits non-zero afterwards.
 
 ### `st tmux`
 
--- a/src/commands/shell_setup.rs
+++ b/src/commands/shell_setup.rs
@@ -602,11 +602,17 @@ fn maybe_install_skills(options: &SetupOptions) -> Result<()> {
     }
 
     println!();
-    skills::run_update_with(
+    if let Err(err) = skills::run_update_with(
         false,
         &HarnessSelection::Only(ids),
         skills::SkillsUpdateOrigin::Setup,
-    )
+    ) {
+        eprintln!(
+            "{}",
+            format!("Warning: some agent skill files could not be written: {err:#}").yellow()
+        );
+    }
+    Ok(())
 }
 
 fn maybe_setup_auth(options: &SetupOptions) -> Result<()> {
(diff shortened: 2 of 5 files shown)


cesarferreira:
## Stack Links

* `main`
  * **PR #868** 👈

